PR #7, opened 2023-02-07
Add initial tooling

stephenjwatkins:
Adds skeletons for the following tools:

- [x] Editorconfig
- [x] Turborepo
- [x] Formatting
- [x] ESLint
- [x] Stylelint
- [x] TypeScript

These are only a starting point and will need to be adjusted as we further implement the library.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "EditorConfig.editorconfig",
+    "esbenp.prettier-vscode",
+    "stylelint.vscode-stylelint"
+  ]
+}
--- a/easy-ui-react/tsconfig.json
+++ b/easy-ui-react/tsconfig.json
@@ -1,0 +1,3 @@
+{
+  "extends": "@easypost/easy-ui-tsconfig/library"
+}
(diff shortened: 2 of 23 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "EditorConfig.editorconfig",
+    "esbenp.prettier-vscode",
+    "stylelint.vscode-stylelint"
+  ]
+}
```
stephenjwatkins, line 8: let me know if these extension recommendations are helpful or just noisy

easy-ui-tsconfig/library.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./dom.json",
+  "compilerOptions": {
+    "declaration": true,
+    "declarationMap": true,
+    "noEmit": false,
+    "noEmitHelpers": true
+  }
+}
```
stephenjwatkins, line 9: admittedly i don't know the nuances of all these rules. they were inspired by configuration from other design systems, so i think it can be a good starting point

.vscode/extensions.json
OskiTheCoder: i think they're quite helpful, we could even at some point later include an example-settings file

easy-ui-react/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "@easypost/easy-ui-tsconfig/library"
```
OskiTheCoder, line 2: really clean, i like this a lot

OskiTheCoder (approved):
🎉

PR #8, opened 2023-02-07
docs: component api best practices

OskiTheCoder:


.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,61 @@
+# Contributing
+
+Please read EasyPost's general [contribution guidelines](https://github.com/EasyPost/.github/blob/main/CONTRIBUTING.md)
```
OskiTheCoder, line 3: does it feel a bit awkward to have to redirect them again to read the general guidelines?

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Booleans
+
+- Booleans should be prefixed with simple verbs. For a button, the property name “isActive” is preferred to “active”.
+- By default booleans should be false. This makes the notation a bit cleaner for when the boolean is active:
+
+  ```JSX
+    // avoid
+   <Button isNotDisabled={false} />
+  ```
```
stephenjwatkins, line 37: looks like there's some extra space at the front of this snippet

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,61 @@
...
+
+Properties that have identical purposes across different components should share the same name. Similarly, properties that have identical purposes should also share the same type. If one component has a boolean “isValid”, other components that intend to control validity should do the same.
+
+### Property dependency
+
+Although unavoidable at times, we should do our best to not create properties that are dependent upon another. This helps avoid configurations that lead to invalid states. If a component has multiple dependent properties, take a closer look at the overall structure of the component.
+
+### Booleans
+
+- Booleans should be prefixed with simple verbs. For a button, the property name “isActive” is preferred to “active”.
```
stephenjwatkins, line 31: for code tokens in the document it may be best to put them in the code tilde:

```suggestion
- Booleans should be prefixed with simple verbs. For a button, the property name `isActive` is preferred to `active`.
```

.github/CONTRIBUTING.md
stephenjwatkins: to me i think it makes sense. company-level guidelines, repo-level guidelines

stephenjwatkins, line 3: are these top two statements sentences? should we end them with a `.`? i'm not a grammar expert lol so i'm curious

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,61 @@
...
+  ```JSX
+  // preferred
+  <Button isDisabled />
+  ```
+
+- Avoid creating booleans with hierarchical structure. If an icon component can have more than two possible placements, avoid “isIconStart”, “isIconMiddle”, and “isIconEnd” since we’d have to decide which variant takes precedence. Instead use an enum.
+
+### Enums
+
+- If an enum has only two values, a boolean should be used instead
```
stephenjwatkins, line 48: i'm not sure i agree with this one in an absolute sense haha. i see them as being slightly semantically different, and i think there are cases where you may want to enumerate only two values. is there a way we could soften the wording to be like "If an enum has only two values, consider if a boolean would make more sense."?

stephenjwatkins, line 48: do we need a `.` at the end? 🤔  not sure what the rule is on those

.github/CONTRIBUTING.md
OskiTheCoder: good idea!

.github/CONTRIBUTING.md
OskiTheCoder: good catch, i think all the sentences here, even the simple sentences should end with a period.

.github/CONTRIBUTING.md
OskiTheCoder: thats a good point and i think im going to remove this part entirely. i got the idea from MUI as they have it in their docs:

Screen Shot 2023-02-08 at 11 23 03 AM

but they also have a couple instances of not really adhering to this and if i remember correctly, react-aria has some cases where they enumerate over two values as well. of course, we can get around those cases, but at that point we'd really be doing it for aesthetics as opposed to a functional purpose. i think as this library matures we can possibly comeback to adding a bit about it with the notable exceptions.

stephenjwatkins: those are great points. that sounds good to me, we can revisit as we better understand how we'd like to do it.

i actually think the second half of the enum statement from MUI sums up my thoughts on it well `or if there is the possibility that additional possible values may be required in the future`. basically how i see it is if the total potential domain is two a boolean makes sense, whereas if the total potential domain is more than 2, even if it currently is 2, likely an enum makes more sense.

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+```JSX
+  // avoid
+  <Button isNotDisabled={false} />
+```
+
+```JSX
+// preferred
+<Button isDisabled />
+```
```
stephenjwatkins, line 42: this is still showing as differently indented, is that intentional?

image

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+- For components that process multiple events with the same verb, they should be named with a descriptive noun in between. For a table component that needs to respond to changes to its column and row, use `onRowChange` and `onColumnChange`.
+
+### Abstractions
+
+- Abstractions are really difficult to get right and reason about. No abstraction is preferred to clunky abstractions. Ideally we want our components to be as composable as possible.
+- For components that need to render a subcomponent, use the slot paradigm of cloning the React element.
+
+### Styling
+
+- Components should not expose the classname name property as we run the risk of creating inconsistent UX.
```
stephenjwatkins, line 57: thoughts on this?

```suggestion
- Components should not expose the `className` property as we run the risk of creating inconsistent UX.
```

stephenjwatkins (commented):
looks good overall, just a couple comments i left

.github/CONTRIBUTING.md
OskiTheCoder: ah no not intentional, will fix!

.github/CONTRIBUTING.md
OskiTheCoder: "classname name" sounded good in my head a few days ago ahahah.

stephenjwatkins (approved):
awesome

PR #9, opened 2023-02-09
Add initial Storybook setup

stephenjwatkins:
Adds initial Storybook setup.

- Add Storybook. This is done differently than the initial thinking of using a separate package. Instead this creates a `.storybook` at the root of the repo to be able to create stories in and for all packages. I don't love that it adds dependency bloat to the root but from my own research it was the best way to go about it, given what we want to accomplish.
- Uses Vite builder which requires adding `vite` to root and basic `vite` config
- Create `<Placeholder />` component to test tech stack
- Modify eslint rules to accommodate TypeScript more properly
- Add `.vscode/settings.json` file to gitignore
- Fix `stylelint` config to use proper config package
- Add CSS/image type definitions

Again some of this will need to be adjusted as we further implement the library.


.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
stephenjwatkins, line 1: this is needed for Storybook

.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<script>
+  window.global = window;
+</script>
```
stephenjwatkins, line 3: this is needed for Vite interactions compatibility

.storybook/preview.js
```diff
@@ -0,0 +1,9 @@
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/,
+    },
+  },
+};
```
stephenjwatkins, line 9: this came default with the Storybook initialization process but seems reasonable to me

package.json
```diff
@@ -23,21 +23,33 @@
...
+    "@babel/core": "^7.20.12",
+    "@storybook/addon-actions": "^6.5.16",
+    "@storybook/addon-essentials": "^6.5.16",
+    "@storybook/addon-interactions": "^6.5.16",
+    "@storybook/addon-links": "^6.5.16",
+    "@storybook/builder-vite": "^0.3.0",
+    "@storybook/react": "^6.5.16",
+    "@storybook/testing-library": "^0.0.13",
+    "@vitejs/plugin-react": "^3.1.0",
+    "babel-loader": "^8.3.0",
```
stephenjwatkins, line 47: unfortunate that we have this bloat in the root package but it seems to be the best place it makes sense for a design system with multiple packages to have potential stories in each

.storybook/main.js
```diff
@@ -0,0 +1,16 @@
+/** @type {import('@storybook/react/types').StorybookConfig} */
+module.exports = {
+  stories: ["../easy-ui-react/src/**/*.stories.mdx"],
+  addons: [
+    "@storybook/addon-links",
+    "@storybook/addon-essentials",
+    "@storybook/addon-interactions",
+  ],
```
OskiTheCoder, line 8: question: do actions need to be included here as well?

suggestion: what are your thoughts on the viewport addon https://storybook.js.org/docs/react/essentials/viewport ?

easy-ui-react/src/components/Placeholder/Placeholder.stories.mdx
```diff
@@ -0,0 +1,17 @@
...
+# Placeholder
+
+With `MDX`, we can define a story for `Placeholder` right in the middle of our
+Markdown documentation.
+
+<Canvas>
+  <Story name="Default" args={{ text: "Test" }}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
```
OskiTheCoder, line 17: MDX looks great, glad we can start using it

did you have a feel for how you wanted to use MDX for easy-ui?

OskiTheCoder (commented):
looking awesome so far, left a couple questions/comments!

easy-ui-react/src/components/Placeholder/Placeholder.stories.mdx
stephenjwatkins: i don't quite have a feel for how to use it yet. i think we'll have to come up with some conventions, but it's nice that it allows for more long-form documentation, which could be beneficial. that said, it does come with some cons that i'd love to talk about with the team

OskiTheCoder (approved):
looks great

PR #12, opened 2023-02-15
Setup base typography

stephenjwatkins:
Adds our base typography configuration to the project.

Based on research, I'm opting currently not to include the actual font assets as part of the installation process. There are a number of ways users could use the Easy UI library that would make it difficult to be made generic. See `easy-ui-react` README for details on what it would look like to set up the project.

- Adds Poppins font for Storybook
- Adds Poppins fallback ligatures
- Adds global font declarations
- Adds global `<EasyUIProvider />` for importing global styles and setting the library up for upcoming global context (Theming, Toasts, etc)


ldewald (approved):
Tested it in storybook and it looks good to me!

OskiTheCoder (approved):
awesome

PR #18, opened 2023-02-27
Add bundling for Easy UI React project

stephenjwatkins:
Adds bundling for Easy UI React.

- Uses Material UI paradigm for organization and build
- Reorganizes components into top-level folders
- Makes `dist/` folder a self-contained package for publishing
- Builds to ESM and CJS
- Exports type declarations


easy-ui-tsconfig/base.json
```diff
@@ -5,7 +5,7 @@
     "isolatedModules": true,
     "experimentalDecorators": true,
     "lib": ["dom", "es2020"],
-    "moduleResolution": "node16",
+    "moduleResolution": "node",
```
stephenjwatkins, line 8: turns out there are some quirks with `node16` that prevent us from using it smoothly currently. TS 5.0 is supposed to address some of these

package.json
```diff
@@ -46,9 +47,9 @@
     "@storybook/builder-vite": "^0.3.0",
     "@storybook/react": "^6.5.16",
     "@storybook/testing-library": "^0.0.13",
-    "@vitejs/plugin-react": "^3.1.0",
```
stephenjwatkins, line 49: moved this to `easy-ui-react` as it's not used directly in a root context

easy-ui-react/tsconfig.json
```diff
@@ -1,8 +1,15 @@
 {
   "extends": "@easypost/easy-ui-tsconfig/library",
   "compilerOptions": {
+    "emitDeclarationOnly": true,
     "baseUrl": "./src",
-    "rootDir": "."
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "paths": {
+      "@easypost/easy-ui-icons/*": ["../../easy-ui-icons/dist/*"],
```
OskiTheCoder, line 9: nice

easy-ui-react/vite.config.ts
```diff
@@ -0,0 +1,58 @@
...
+    return { ...o, [alias]: entry };
+  }, {});
+}
+
+// Cleans up package.json file for including in distribution
+function transformPkgJson(code) {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { scripts, devDependencies, exports, ...restPkg } = JSON.parse(code);
+  return JSON.stringify(restPkg, null, 2);
+}
```
OskiTheCoder, line 58: out of curiosity, why was the config file moved into easy-ui-react?

stephenjwatkins: i originally created it in the root because i thought Storybook used it, but it actually doesn't lol. it uses our Vite instance but it has its own config (by default). so then i realized only our easy-ui-react project is using a Vite config so i just moved it in there for now. it's possible we'll have other projects in the future that we'll warrant us sharing config across projects in which case it'd probably make sense to have a root one at that point

OskiTheCoder (approved):
lgtm!

PR #19, opened 2023-02-28
basic ci setup

OskiTheCoder:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Install dependencies
+        run: npm i
+      - name: Build tokens
+        run: npm run build:tokens
+      - name: Build icons
+        run: npm run build:icons
+      - name: Formatting
+        run: npm run format
+      - name: Unit tests
+        run: npm run test
```
OskiTheCoder, line 27: i didn't see that we had a script for linting, but i could look into adding it

stephenjwatkins: we do need to lint in CI. you could either add it in this task or create a separate task for it. up to you.

it would essentially involve having a `lint` command on a per-project basis that runs appropriate linters for that project (eslint on TS/JS files and stylelint on (S)CSS files). then having a `lint` command at the root that runs the project `lint` commands via turbo

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          node-version: "16.x"
+      - name: Install dependencies
+        run: npm i
+      - name: Build tokens
+        run: npm run build:tokens
+      - name: Build icons
+        run: npm run build:icons
+      - name: Formatting
+        run: npm run format
```
stephenjwatkins, line 25: `npm run format` actually formats the files and writes them to disk. in CI we want to actually just run a check on the format. so we want another script like `format:check` that runs `prettier . --check` instead of `prettier . --write`

stephenjwatkins (changes requested):
looks good, just want to figure out direction on lint and then fix the prettier format check

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request: ~
+  workflow_dispatch: ~
```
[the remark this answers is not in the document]
OskiTheCoder: same haha, and yep your guess aligns with my understanding too. if we just had push without a branch specifier, it would run on push and when we create the PR and i figured it would be ok if we just ran it on just PR. similarly, i guess if we _need_ to ever force push to main, we could still have some CI. as for the workflow dispatch, i checked with the dev ex team and they've had some instances where they found it useful to be able to manually run a CI.

easy-ui-react/.stylelintrc
```diff
@@ -0,0 +1,13 @@
+{
+  "root": true,
+  "extends": ["@easypost/stylelint-easy-ui"],
+  "overrides": [
+    {
+      "files": ["src/**/*.scss"],
+      "customSyntax": "postcss-scss"
```
OskiTheCoder, line 7: when i tried to run the stylelint without this, i received this console message 

Screen Shot 2023-03-01 at 3 17 37 PM

admittedly im not super familiar with it, but i went ahead and added `stylelint-config-recommended-scss` and created a separate override. although again, not totally sure if this is the ideal way

.github/workflows/ci.yml
OskiTheCoder: derp, missed that hah

easy-ui-react/.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+  "root": true,
+  "extends": ["@easypost/eslint-config-easy-ui"],
+  "overrides": [
+    {
+      "files": ["src/components/**/*.tsx"],
```
OskiTheCoder, line 6: should this include every tsx/ts file in `easy-ui-react`?

stephenjwatkins: actually i forgot that i modified the main eslint config as part of the responsive props tasks that i think fixes issues you were running into?

the eslint config for the repo is governed by the `eslint-config-easy-ui` repo. in `eslint-config-easy-ui/index.js` i made these changes. with them, you shouldn't need an `.eslintrc` in `easy-ui-react`:

image

```js
/** @type {import('eslint').Linter.Config} */
module.exports = {
  extends: [
    "turbo",
    "plugin:react/recommended",
    "plugin:react-hooks/recommended",
    "plugin:storybook/recommended",
    "plugin:prettier/recommended",
  ],
  settings: {
    react: {
      version: "detect",
    },
  },
  env: {
    node: true,
    browser: true,
  },
  overrides: [
    {
      files: ["**/*.js", "**/*.jsx"],
      extends: ["eslint:recommended"],
      parserOptions: {
        ecmaVersion: "2021",
        sourceType: "module",
      },
    },
    {
      files: ["**/*.ts", "**/*.tsx"],
      extends: ["plugin:@typescript-eslint/recommended"],
      plugins: ["@typescript-eslint"],
      parser: "@typescript-eslint/parser",
    },
  ],
};
```

easy-ui-react/.stylelintrc
stephenjwatkins: ohh thanks for the heads up on that. forgot about modifying stylelint for scss. in that case, i think there are a couple things we need to do:

in `stylelint-easy-ui/package.json` uninstall (with `npm uninstall` so it saves to `package-lock.json`) the regular `recommended` package:

image

in `stylelint-easy-ui/index.js` replace the regular `recommended` package with the scss one:

image

and with that package in place, it'll complain about including the `.scss` extension inside of a partial, so we need to remove that in `easy-ui-react/src/styles/_common.scss`:

image

then with this in place for the whole project, we can remove `easy-ui-react/.stylelintrc`

easy-ui-react/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@easypost/easy-ui",
   "version": "1.0.0",
   "scripts": {
+    "lint": "eslint src --ext .ts,.tsx && stylelint 'src/**/*.{css,scss}' && npx prettier --check ./src",
```
stephenjwatkins, line 5: you know what, i forgot that our eslint is configured to check prettier as well so we can remove the prettier check line :)

then i think we can optimize this lint script with the following:

image

that way they can take advantage of being run in parallel (they don't depend on each other). `run-p` uses the `npm-run-all` package that's installed at the root to run tasks in parallel

package.json
```diff
@@ -31,6 +31,7 @@
   "scripts": {
     "clean": "FORCE_COLOR=1 turbo run clean && rm -rf ./node_modules",
     "format": "prettier . --write",
+    "lint": "turbo run lint",
```
stephenjwatkins, line 34: for the time being, we need to prepend turbo scripts with `FORCE_COLOR=1` to preserve underlying color. hoping they fix this soon

```suggestion
    "lint": "FORCE_COLOR=1 turbo run lint",
```

stephenjwatkins (changes requested):
looking good. i think i have the final requests outlined

stephenjwatkins (approved):
awesome

PR #21, opened 2023-03-03
Add Icon component

stephenjwatkins:
Adds Icon component per spec.

- Add <Icon />
- Add support for component tokens


.storybook/preview.jsx
```diff
@@ -7,7 +7,6 @@ export const parameters = {
   actions: { argTypesRegex: "^on[A-Z].*" },
   controls: {
     matchers: {
-      color: /(background|color)$/i,
```
stephenjwatkins, line 10: this ended up colliding with us wanting to limit colors to our own tokens, so needed to remove this

OskiTheCoder (approved):
awesome

OskiTheCoder:
we could probably also remove the `Placeholder` component unless we need it for a bit longer

PR #28, opened 2023-03-08
Improve Storybook configuration

stephenjwatkins:
- Removes unused addons
- Adds accessibility addon
- Brands the Storybook window for EasyPost
- Uses Easy UI breakpoints
- Uses Easy UI grid size
- Update gray tokens to match Figma names (whoops!)


.storybook/main.js
```diff
@@ -1,11 +1,7 @@
 /** @type {import('@storybook/react/types').StorybookConfig} */
 module.exports = {
   stories: ["../easy-ui-react/src/**/*.stories.mdx"],
-  addons: [
-    "@storybook/addon-links",
-    "@storybook/addon-essentials",
-    "@storybook/addon-interactions",
-  ],
+  addons: ["@storybook/addon-essentials", "@storybook/addon-a11y"],
```
stephenjwatkins, line 4: if in the future if we find we want the other addons we can add them them. didn't want to pollute our addons with unused ones

easy-ui-tokens/src/color/base.json
```diff
@@ -51,20 +51,20 @@
...
+      "800": { "value": "#1C222D" },
+      "700": { "value": "#384359" },
+      "600": { "value": "#47547F" },
+      "500": { "value": "#7086B2" },
+      "400": { "value": "#8D9EC1" },
+      "300": { "value": "#A9B6D1" },
+      "200": { "value": "#C6CFE0" },
+      "100": { "value": "#E2E7F0" },
+      "050": { "value": "#F1F1F1" },
       "000": { "value": "#FFFFFF" }
```
stephenjwatkins, line 64: didn't realize they were named particularly in Figma. whoops

.storybook/theme.js
```diff
@@ -0,0 +1,55 @@
...
+  barTextColor: colors.text,
+  barSelectedColor: colors.primary,
+  barBg: colors.background,
+
+  // Form colors
+  inputBg: colors.background,
+  inputBorder: colors.border,
+  inputTextColor: colors.text,
+  inputBorderRadius: tokens["shape-border-radius-base"],
+});
```
stephenjwatkins, line 55: we can change this to be something different in the future per design recs, just wanted to get a structure in place for some kind of identity

ldewald (approved):
Approving as is but a couple questions for follow up:
* Is it possible to default to the `Docs` view rather than `Canvas`? 
* Is it possible to set the favicon?

PR #29, opened 2023-03-09
button

OskiTheCoder:


easy-ui-react/src/Button/Button.tsx
```diff
@@ -0,0 +1,86 @@
...
+import styles from "./Button.module.scss";
+
+export type ButtonColor =
+  | "primary"
+  | "secondary"
+  | "success"
+  | "warning"
+  | "neutral"
+  | "support"
+  | "inverse";
```
OskiTheCoder, line 15: i went back and forth on this a lot. the issues i ran into when trying to come up with a good representation for different colors is that not all of our outlined variations have filled counterparts and its not particularly obvious how to handle invalid states. i did think about just dumping all of the variations into a single property, something like 

`ButtonColorStyles = “primary_filled” | “primary_outlined” | “secondary_filled”, | “secondary_outlined” ..etc` 

and im still open to it, but it does deviate quite a bit from how we’ve been doing it. 

with the way i implemented it here, we do still have some invalid states, but handling that under this structure would mean detecting a fallback which im not sure how friendly that is to the dx.

curious if you have thoughts and im open to ideas.

easy-ui-react/src/Button/Button.tsx
```diff
@@ -0,0 +1,86 @@
...
+  } = props;
+
+  const ref = useRef(null);
+  const As = href ? "a" : "button";
+  const { buttonProps: elementProps } = useButton(
+    { ...props, elementType: As },
+    ref,
+  );
+  const canUseIcon =
+    (startIcon || endIcon) && appearance !== "link" && size !== "sm";
```
OskiTheCoder, line 60: no icons for small buttons and the link variation per design

easy-ui-react/src/styles/global.scss
```diff
@@ -22,3 +22,8 @@ select,
 textarea {
   font-family: inherit;
 }
+
+button,
+a {
+  cursor: pointer;
+}
```
OskiTheCoder, line 29: does this make sense as a global style?

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  &.sizeSmFilled {
+    padding: var(--ezui-space-1);
+  }
+
+  &.sizeSmOutlined,
+  &.sizeSmLink {
+    padding: calc(#{var(--ezui-space-1)} - #{var(--ezui-shape-border-width-1)});
+  }
```
OskiTheCoder, line 271: will double check with design about the height for the small button, i presume its intended to be 32px, but the figma has 36px due to the previous line height

easy-ui-react/src/Button/Button.stories.mdx
```diff
@@ -0,0 +1,367 @@
...
+    name="ClickEvent"
+    args={{
+      onPress: () => alert("clicked!"),
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable story="Default" />
```
OskiTheCoder, line 367: i tried to write a function that would generate these stories, but i was getting some strange type errors, might revisit this though because it does feel a little heavy to have a bunch of stories manually written out

easy-ui-react/src/Button/Button.tsx
stephenjwatkins: i like what you decided on

if there are invalid variant combinations, we can log a warning: https://github.com/Shopify/polaris/blob/main/polaris-react/src/components/Icon/Icon.tsx#L60

easy-ui-react/src/Button/Button.tsx
```diff
@@ -0,0 +1,86 @@
...
+  | "support"
+  | "inverse";
+export type ButtonAppearance = "filled" | "outlined" | "link";
+export type ButtonSize = "sm" | "md";
+
+export type ButtonProps = {
+  /** Button color */
+  color?: ButtonColor;
+  /** Button appearance */
+  appearance?: ButtonAppearance;
```
stephenjwatkins, line 23: how would you feel about using `variant` instead of `appearance`? i know we used `appearance` previously but from research it seems `variant` is a little bit more ubiquitous/versatile in component libraries. hoping we can standardize on a convention to use throughout

easy-ui-react/src/Button/Button.tsx
```diff
@@ -0,0 +1,86 @@
...
+  /** Button size */
+  size?: ButtonSize;
+  /** Disables button */
+  isDisabled?: boolean;
+  /** Button will grow to width of container */
+  isBlock?: boolean;
+  /** Positions icon before children */
+  startIcon?: ReactElement<IconProps>;
+  /** Positions icon after children */
+  endIcon?: ReactElement<IconProps>;
```
stephenjwatkins, line 33: i've seen `iconAtStart` and `iconAtEnd` as names recently in other libraries and i'm wondering if that's slightly better? 🤔

easy-ui-tokens/src/font/base.json
```diff
@@ -16,6 +16,7 @@
       "50": { "value": "12px", "attributes": { "unit": "rem" } },
       "75": { "value": "13px", "attributes": { "unit": "rem" } },
       "100": { "value": "15px", "attributes": { "unit": "rem" } },
+      "125": { "value": "16px", "attributes": { "unit": "rem" } },
```
stephenjwatkins, line 19: i feel a little uneasy about the button text being 1px bigger than the body text 🤔 did we confirm that that's indeed what design wants?

easy-ui-react/src/Button/Button.stories.mdx
```diff
@@ -0,0 +1,367 @@
...
+  <Story name="Default" args={{}}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+Filled buttons
+
+<Canvas>
+  <Story
+    name="Secondary"
```
stephenjwatkins, line 34: is it possible to organize these stories under certain groupings to clean up the sidebar?

`name="Filled/Secondary"`

can't remember if it supports nesting at this level

easy-ui-react/src/Button/Button.stories.mdx
```diff
@@ -0,0 +1,367 @@
...
+
+# Button
+
+<Canvas>
+  <Story name="Default" args={{}}>
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+Filled buttons
```
stephenjwatkins, line 30: would it make sense to use an H2?

`## Filled buttons`

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,285 @@
...
+  &.colorPrimaryFilled {
+    @include filled-button(
+      var(--ezui-color-white),
+      var(--ezui-color-white),
+      var(--ezui-color-blue-500),
+      var(--ezui-color-blue-600),
+      var(--ezui-color-blue-700),
+      var(--ezui-shadow-button)
+    );
+  }
```
stephenjwatkins, line 122: in CSS modules i don't believe we need to nest it due to how modules isolates it already to the component. could help slightly IMO with cognitive overhead

stephenjwatkins, line 122: is there a way we could separate the individual variants into their own classes, keeping them separate from each other? the way it is right now is going to generate quite a lot of CSS and make it a little less portable.

```
// set colors
.colorPrimary {
  color: color;
  border-color: color;
  background: color;
}

// clear out unused properties, etc
.variantOutline {
  background: none;
}

.variantFilled {
  border: 0;
}
```

that way it'll be easier to keep color and structure easier to maintain. let me know if that makes sense

stephenjwatkins (commented):
looks awesome

easy-ui-react/src/styles/global.scss
stephenjwatkins: i would probably try to isolate css changes to our components where possible, only because global styles can conflict with external libraries and such. if we want EZUI to work anywhere, keeping it encapsulated per component would be best

PR #31, opened 2023-03-13
put button pr backup

OskiTheCoder:
old PR https://github.com/EasyPost/easy-ui/pull/29

### DIFF
--- a/easy-ui-react/package.json
+++ b/easy-ui-react/package.json
@@ -20,7 +20,8 @@
   ],
   "dependencies": {
     "@types/react": "^18.0.27",
-    "@types/react-dom": "^18.0.10"
+    "@types/react-dom": "^18.0.10",
+    "react-aria": "^3.23.1"
   },
   "peerDependencies": {
     "react": "^16.14.0 || ^17.0.0 || ^18.0.0",
@@ -33,7 +34,6 @@
     "@vitejs/plugin-react": "^3.1.0",
     "jsdom": "^21.1.0",
     "react": "^18.2.0",
-    "react-aria": "^3.23.1",
     "react-dom": "^18.2.0",
     "sass": "^1.58.3",
     "vite-plugin-static-copy": "^0.13.1"
--- a/easy-ui-react/package.json
+++ b/easy-ui-react/package.json
@@ -33,6 +33,7 @@
     "@vitejs/plugin-react": "^3.1.0",
     "jsdom": "^21.1.0",
     "react": "^18.2.0",
+    "react-aria": "^3.23.1",
     "react-dom": "^18.2.0",
     "sass": "^1.58.3",
     "vite-plugin-static-copy": "^0.13.1"
--- a/easy-ui-react/src/Button/Button.module.scss
+++ b/easy-ui-react/src/Button/Button.module.scss
@@ -1,0 +1,321 @@
+@import "../styles/common";
+
+.Button {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  border-radius: var(--ezui-shape-border-radius-base);
+  cursor: pointer;
+
+  &:focus {
+    box-shadow: var(--ezui-shadow-button);
+    outline: none;
+  }
+
+  &:disabled {
+    pointer-events: none;
+  }
+}
+
+.colorPrimary {
+  color: var(--ezui-color-blue-500);
+  border-color: var(--ezui-color-blue-500);
+  background-color: var(--ezui-color-blue-500);
+}
+
+.colorSecondary {
+  color: var(--ezui-color-purple-500);
+  border-color: var(--ezui-color-purple-500);
+  background-color: var(--ezui-color-purple-500);
+}
+
+.colorSuccess {
+  color: var(--ezui-color-green-500);
+  border-color: var(--ezui-color-green-500);
+  background-color: var(--ezui-color-green-500);
+}
+
+.colorWarning {
+  color: var(--ezui-color-red-500);
+  border-color: var(--ezui-color-red-500);
+  background-color: var(--ezui-color-red-500);
+}
+
+.colorNeutral {
+  color: var(--ezui-color-white);
+  border-color: var(--ezui-color-blue-800);
+  background-color: var(--ezui-color-blue-800);
+}
+
+.colorSupport {
+  color: var(--ezui-color-blue-800);
+  border-color: var(--ezui-color-blue-800);
+}
+
+.colorInverse {
+  color: var(--ezui-color-white);
+  border-color: var(--ezui-color-white);
+}
+
+.variantOutlined {
+  background-color: transparent;
+  border-style: solid;
+  border-width: var(--ezui-shape-border-width-1);
+
+  &:link {
+    text-decoration: none;
+  }
+
+  &:disabled {
+    color: var(--ezui-color-gray-300);
+    border-color: var(--ezui-color-gray-300);
+  }
+  &:active:not(:disabled) {
+    &:focus {
+      box-shadow: none;
+    }
+  }
+}
+
+.variantFilled {
+  border: 0;
+
+  &:link {
+    text-decoration: none;
+  }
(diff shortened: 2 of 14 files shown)


easy-ui-react/package.json
```diff
@@ -33,6 +33,7 @@
     "@vitejs/plugin-react": "^3.1.0",
     "jsdom": "^21.1.0",
     "react": "^18.2.0",
+    "react-aria": "^3.23.1",
```
OskiTheCoder, line 36: is this okay as a dev dep or should it be a standard dep?

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,321 @@
...
+    box-shadow: var(--ezui-shadow-button);
+    outline: none;
+  }
+
+  &:disabled {
+    pointer-events: none;
+  }
+}
+
+.colorPrimary {
```
OskiTheCoder, line 20: addressed feedback for styles, let me know if i misunderstood the suggestion

easy-ui-react/src/Button/Button.stories.mdx
```diff
@@ -0,0 +1,157 @@
...
+  gap: "12px",
+  flexWrap: "wrap",
+};
+
+export const inverseBackground = {
+  backgroundColor: "var(--ezui-color-blue-800)",
+  padding: "12px",
+};
+
+# Button
```
OskiTheCoder, line 34: cleaned up button stories, i wasn't able to nest at the story level, but open to suggestions about a better grouping

easy-ui-react/src/Button/Button.tsx
```diff
@@ -0,0 +1,113 @@
...
+  | "support"
+  | "inverse";
+export type ButtonVariant = "filled" | "outlined" | "link";
+export type ButtonSize = "sm" | "md";
+
+export type ButtonProps = {
+  /** Button color */
+  color?: ButtonColor;
+  /** Button variant */
+  variant?: ButtonVariant;
```
OskiTheCoder, line 23: updated to variant

easy-ui-tokens/src/font/base.json
```diff
@@ -16,6 +16,7 @@
       "50": { "value": "12px", "attributes": { "unit": "rem" } },
       "75": { "value": "13px", "attributes": { "unit": "rem" } },
       "100": { "value": "15px", "attributes": { "unit": "rem" } },
+      "125": { "value": "16px", "attributes": { "unit": "rem" } },
```
OskiTheCoder, line 19: pending feedback about text styles

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,321 @@
...
+}
+
+.sizeMd.variantOutlined,
+.sizeMd.variantLink {
+  padding: calc(#{var(--ezui-space-1-5)} - #{var(--ezui-shape-border-width-1)})
+    calc(#{var(--ezui-space-2)} - #{var(--ezui-shape-border-width-1)});
+}
+
+.sizeSm.variantFilled {
+  padding: var(--ezui-space-1);
```
OskiTheCoder, line 302: pending feedback with regards to small button height

easy-ui-react/package.json
stephenjwatkins: it should actually be a standard dep since it'll be "included" in the lib. in contrast, react is a peer dependency that's excluded from the build, for instance. we'll also want to make sure Vite is not including all of react-aria during build. you can test this by running `npm run build:react` and looking at the file sizes that are output. if it's huge, it's likely including the entire library instead of tree shaking it

OskiTheCoder: added as regular dep, i compared the size of `dist` when using `react-aria` vs `@react-aria/button` and they appear to be the same, i presume that means the library was correctly tree shaken

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,321 @@
...
+
+.sizeMd.variantFilled,
+.sizeMd.variantOutlined,
+.sizeMd.variantLink {
+  @include font-style("button");
+}
+
+.sizeSm.variantFilled,
+.sizeSm.variantOutlined,
+.sizeSm.variantLink {
```
stephenjwatkins, line 287: can these just be `.sizeMd` and `.sizeSm`?

easy-ui-react/src/Button/Button.module.scss
stephenjwatkins: yep you understood the suggestion. it does lead to a lot of CSS, a bit more than i realized it would. i think we can clean this up a bit with component-level CSS vars like we discussed in Slack

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,321 @@
...
+  width: 100%;
+}
+
+.StartIcon {
+  margin-left: var(--ezui-space-1);
+}
+
+.EndIcon {
+  margin-right: var(--ezui-space-1);
+}
```
stephenjwatkins, line 321: with CSS modules, i tend to see classes written in more typical CSS syntax and without the root modifier. in this case, i think these could become:

```
.block {}

.iconAtStart {}

.iconAtEnd {}
```

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,321 @@
...
+
+.sizeSm.variantOutlined,
+.sizeSm.variantLink {
+  padding: calc(#{var(--ezui-space-1)} - #{var(--ezui-shape-border-width-1)});
+}
+
+.ButtonBlock {
+  display: flex;
+  width: 100%;
```
stephenjwatkins, line 312: not that this is wrong, but out of curiosity, what's the reasoning for using `display: flex; width: 100%;` instead of `display: block;`?

easy-ui-react/src/Button/Button.stories.mdx
stephenjwatkins: i find this update much better

easy-ui-react/src/Button/Button.module.scss
stephenjwatkins: oh nvm duh

easy-ui-react/src/styles/_common-button.scss
```diff
@@ -0,0 +1,119 @@
...
+  --ezui-c-button-filled-font-color: var(--ezui-color-white);
+  --ezui-c-button-filled-active-font-color: var(--ezui-color-white);
+  --ezui-c-button-resting-color: var(--ezui-color-purple-500);
+  --ezui-c-button-hover-focus-color: var(--ezui-color-purple-800);
+  --ezui-c-button-active-color: var(--ezui-color-purple-700);
+}
+
+@mixin button-success-color-aliases() {
+  --ezui-c-button-filled-font-color: var(--ezui-color-blue-800);
+  --ezui-c-button-filled-active-font-color: var(--ezui-color-white);
```
OskiTheCoder, line 19: success color has a different text color when in the active state

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -0,0 +1,119 @@
...
+    color: var(--ezui-c-button-active-color);
+  }
+
+  &:hover:not(:disabled):not(:active) {
+    color: var(--ezui-c-button-hover-focus-color);
+  }
+}
+
+.sizeMd {
+  @include font-style("button");
```
OskiTheCoder, line 82: checked with design and we do indeed want 15px for the button text, removed changes to font

easy-ui-tokens/src/font/base.json
OskiTheCoder: 15px is what we want here, not 16px. reverted change to fonts

stephenjwatkins (approved):
this is very good

PR #35, opened 2023-03-22
icon button

OskiTheCoder:
- adds `IconButton` component
- updates styles in `Button` and `Dropdown` to accommodate padding for `IconButton`


easy-ui-react/src/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export type IconButtonVariant = "filled" | "outlined";
+
+export type IconButtonProps = {
+  /** Button color */
+  color?: ButtonColor;
+  /** Button variant */
+  variant?: IconButtonVariant;
+  /** Icon symbol SVG source from @easypost/easy-ui-icons */
+  symbol: IconSymbol;
```
OskiTheCoder, line 18: does this make sense as a symbol instead of the icon component itself? if so im wondering if i should update the button to also accept a symbol

easy-ui-react/src/IconButton/IconButton.module.scss
```diff
@@ -0,0 +1,26 @@
...
+  }
+}
+
+.IconButton.variantOutlined {
+  padding: calc(#{var(--ezui-space-1-5)} - #{var(--ezui-shape-border-width-1)});
+}
+
+.IconButton.variantFilled {
+  padding: var(--ezui-space-1-5);
+}
```
OskiTheCoder, line 26: `IconButton` needs `12px` of padding horizontally, which is a bit different than the `Button` and `DropdownButton` so i moved the padding out of the variant selector. i think another way that works is to have  just have the `IconButton` override whatever is in `Button` but i think it kinda gets tricky when styles are overrided across files

easy-ui-react/src/IconButton/IconButton.module.scss
```diff
@@ -0,0 +1,26 @@
...
+
+  &:focus {
+    box-shadow: var(--ezui-shadow-button);
+    outline: none;
+  }
+
+  &:disabled {
+    pointer-events: none;
+  }
+}
```
stephenjwatkins, line 18: **thought:** i notice that most of these styles are used across the buttons. would it make sense to create a mixin for the shared root styles in `Button.module.scss` to `@use` in `DropdownButton.module.scss` and `IconButton.module.scss`, similar to how we would create mixins for the variants and colors?

easy-ui-react/src/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+      ref={ref}
+      className={classNames(
+        styles.IconButton,
+        styles[variationName("variant", variant)],
+        commonButtonStyles[variationName("color", color)],
+        commonButtonStyles[variationName("variant", variant)],
+      )}
+      {...buttonProps}
+    >
+      {React.cloneElement(icon)}
```
stephenjwatkins, line 48: **issue:** i didn't catch this pattern in the original `Button` pull request, but i believe it would be better to accept a symbol as the prop instead of cloning an `<Icon />` component.

in addition, i would suggest using an accessibility pattern directly within the `<button />` instead of using the underlying `<Icon accessibilityLabel="" />` due to it being marginally more robust across screen readers: https://www.sarasoueidan.com/blog/accessible-icon-buttons/#there-is-no-one-way-to-rule-them-all:~:text=Personally%2C%20I%E2%80%99d%20not,on%20his%20findings.

> Personally, I’d not use the SVG icon itself to provide a label for the button when I can provide one on the button itself directly. As Scott states, there’s just no good way to use an SVG as the sole means to provide an accessible name to a link or button. Every test had failures. Thankfully, Scott has saved us hours of testing, so if you need to provide a label to the button using the svg itsef, you can choose the pattern you need based on his findings.

example of changes:

```tsx
import type { IconSymbol } from "../types";

export type IconButtonProps = {
  /** Icon symbol */
  symbol: IconSymbol;
  /** Description of icon */
  accessibilityLabel: string;
};

export function IconButton(props: IconButtonProps) {
  const {
    symbol,
    accessibilityLabel,
  } = props;
  return (
    <button {...buttonProps}>
      <Text visuallyHidden>{accessibilityLabel}</Text>
      <Icon symbol={symbol} />
    </button>
  );
}

// then use like
<IconButton
  symbol={BackIcon}
  accessibilityLabel="Back"
/>
```

with this paradigm, it allows the `IconButton` component to control the rendering of the icon. with cloning, the user has too much flexibility in changing the rendering of the icon within the button to the point where it would be easy to break. for example, the user could break the size and color scheme with the following: `<IconButton icon={<Icon size="lg" color="blue-800" />} />`

i would also highly suggest the same pattern with the `icon*` props in the original `Button` component for the same reasons, but that could be a follow-up task

stephenjwatkins (changes requested):
looking good, a couple thoughts/suggestions

easy-ui-react/src/IconButton/IconButton.tsx
OskiTheCoder: very interesting, thanks for the info! updated `IconButton` and `Button` . for `Button` i omitted the `Text` component since i imagine the accessible text is given by the `Button` text itself

easy-ui-react/src/IconButton/IconButton.module.scss
OskiTheCoder: agree!

easy-ui-react/src/Button/utilities.ts
```diff
@@ -1,10 +1,11 @@
 import { ButtonColor } from "./Button";
 import { ButtonVariant } from "./Button";
 import { DropdownButtonVariant } from "../DropdownButton";
+import { IconButtonVariant } from "../IconButton";
 
 export function logWarningIfInvalidColorVariantCombination(
   color: ButtonColor,
-  variant: ButtonVariant | DropdownButtonVariant,
+  variant: ButtonVariant | DropdownButtonVariant | IconButtonVariant,
```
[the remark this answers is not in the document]
OskiTheCoder: updated to only use `ButtonVariant`

easy-ui-react/src/IconButton/IconButton.module.scss
OskiTheCoder: really like this approach, updated!

easy-ui-react/src/Button/Button.tsx
```diff
@@ -58,8 +59,20 @@ export function Button(props: ButtonProps) {
     { ...props, elementType: As },
     ref,
   );
+
+  const bothIconPropsDefined = iconSymbolAtEnd && iconSymbolAtStart;
```
OskiTheCoder, line 63: checked with design, no use cases for defining both icons

easy-ui-react/src/Button/Button.stories.mdx
```diff
@@ -72,13 +80,9 @@ export const ButtonStoryOnDarkBackgroundDecorator = (story) => (
   <Story name="OutlinedButtons" decorators={[ButtonStoryDecorator]}>
     <Button variant="outlined" />
     <Button color="secondary" variant="outlined" />
-    <Button color="success" variant="outlined" />
-    <Button color="warning" variant="outlined" />
```
OskiTheCoder, line 76: check with design, no use cases for these combinations and safe to consider as an invalid combination

easy-ui-react/src/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export type IconButtonVariant = "filled" | "outlined";
+
+export type IconButtonProps = {
+  /** Button color */
+  color?: ButtonColor;
+  /** Button variant */
+  variant?: IconButtonVariant;
+  /** Icon symbol */
+  iconSymbol: IconSymbol;
```
stephenjwatkins, line 19: **thought:** based on other comment, curious about changing this property name to just `icon`?

easy-ui-react/src/Button/Button.module.scss
stephenjwatkins, line 1: **note:** we talked on Slack about reorganizing this (again). i didn't realize SASS' `@use` includes classes from other files in its output, at least in a CSS Modules setup. that could create some confusion collision, so we're going to split out mixins into its own file

easy-ui-react/src/Button/Button.tsx
```diff
@@ -29,10 +30,10 @@ export type ButtonProps = {
   isDisabled?: boolean;
   /** Button will grow to width of container */
   isBlock?: boolean;
-  /** Positions icon before children */
-  iconAtStart?: ReactElement<IconProps>;
-  /** Positions icon after children */
-  iconAtEnd?: ReactElement<IconProps>;
+  /** Positions icon symbol before children */
+  iconSymbolAtStart?: IconSymbol;
```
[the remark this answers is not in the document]
OskiTheCoder: ah yeah i meant to add a comment to ask about this, i do much prefer not using `symbol` in the prop name, agree with the suggestion

documentation/specs/Button.md
```diff
@@ -13,31 +13,31 @@ export type ButtonVariant = "filled" | "outlined" | "link";
 export type ButtonSize = "sm" | "md";
 export type ButtonProps = {
   /** Button color */
-  // Defaults to primary
+  /** @default "primary" */
```
stephenjwatkins, line 16: **note:** just an fyi that TSDoc/JSDoc should be a single comment block and not split up across blocks. for example, before:

```tsx
/** Button color */
/** @default "primary" */
color?: ButtonColor;
```
this will only show `@default "primary"` in IDE intellisense.

after:

```tsx
/**
 * Button color
 * @default "primary"
 */
color?: ButtonColor;
```

this will properly show both the summary and `@default` in intellisense

easy-ui-react/src/Button/Button.tsx
```diff
@@ -58,8 +59,18 @@ export function Button(props: ButtonProps) {
...
+  const bothIconPropsDefined = iconAtEnd && iconAtStart;
+  if (bothIconPropsDefined) {
+    // eslint-disable-next-line no-console
+    console.warn("Cannot simultaneously define `iconAtEnd` and `iconAtStart`");
+  }
+
   const canUseIcon =
-    (iconAtEnd || iconAtStart) && variant !== "link" && size !== "sm";
+    (iconAtEnd || iconAtStart) &&
+    !bothIconPropsDefined &&
```
stephenjwatkins, line 71: **nitpick:** since `bothIconPropsDefined` triggers an error above, you could remove this from the check, but not a huge deal either way

PR #79, opened 2023-04-04
Use theme-aware variables

stephenjwatkins:
- Updates CSS files to use theme-aware variables
- Added `theme-token` and `design-token` helpers to keep naming consistent across project
- Updates token name usage to use dot notation to be consistent with how they're defined in the tokens project and theme configuration
- Updates CSS files to use token helpers
- Added `dev` scripts that automatically rebuild on changes

Given that we have different token types now, this change tries to make it easier to work with tokens. Tokens can be one of either (a) design token; (b) theme token; (c) component token. This maps to the original plan of global, alias, and component tokens in our ADR -> design token (global), theme token (alias), and component token.

Design tokens are any tokens in our `-tokens` project, theme tokens are ones defined in our theme in `-react`, component tokens are tokens created in our component CSS module files.


easy-ui-tokens/src/theme/light.json
```diff
@@ -1,24 +1,37 @@
 {
   "theme.light": {
```
stephenjwatkins, line 2: the only thing i don't feel great about with the theming work is having the theme tokens defined both here and in the `Theme.tsx` file.

the `Theme.tsx` file is the source of truth for the theme tokens in our Easy UI project, and the tokens defined here our the specific implementation of that truth for EasyPost branding, but it still feels a little clunky having both. it may be simpler to use the tokens defined here as the truth instead, but i'm not sure if we can make that type safe. i may look into that.

any ideas are welcome.

stephenjwatkins: looks like it is in fact possible to derive the theme definition types from the tokens. i think i like this approach better

OskiTheCoder: agree

easy-ui-react/src/Button/_mixins.scss
```diff
@@ -17,144 +19,152 @@
...
 }
 
+// prettier-ignore
 @mixin colorPrimary {
-  --ezui-c-button-filled-font-color: var(--ezui-color-white);
-  --ezui-c-button-filled-active-font-color: var(--ezui-color-white);
-  --ezui-c-button-resting-color: var(--ezui-color-blue-500);
-  --ezui-c-button-hover-focus-color: var(--ezui-color-blue-600);
-  --ezui-c-button-active-color: var(--ezui-color-blue-700);
+  @include component-token("button", "filled.font.color", theme-token("color.text.inverse"));
```
OskiTheCoder, line 115: really like the use of component tokens here

package.json
```diff
@@ -39,6 +39,7 @@
     "changes:status": "changeset status",
     "changes:version": "changeset version",
     "clean": "FORCE_COLOR=1 turbo run clean && rm -rf ./node_modules",
+    "dev": "FORCE_COLOR=1 turbo run dev",
```
OskiTheCoder, line 42: nice

easy-ui-react/src/utilities/css.ts
```diff
@@ -33,7 +34,23 @@ export function getComponentToken(
...
 
-export function getResponsiveToken(
+export function getComponentThemeToken(
+  componentName: string,
+  componentProp: string,
+  tokenSubgroup: string,
+  token?: string,
+) {
+  if (!token) {
+    return {};
```
OskiTheCoder, line 44: i know this bit isn't necessarily the focus of whats going on here, but out of curiosity, whats the thinking behind making token optional?

stephenjwatkins: i copied the original function from Polaris which did it this way, but i think the idea is that it allows for optional props in a component API can just be fed directly into this function without having to guard against it in the component itself

OskiTheCoder (approved):
awesome!!

PR #82, opened 2023-04-05
Create component spec template

stephenjwatkins:
Here's one thought for a component spec template, based on Open UI's component spec template.

I don't want the spec process to be too heavy handed but enough for us to think through potential issues before implementation.

[Open UI Component Spec Template](https://github.com/openui/open-ui/blob/main/site/src/pages/component-spec-template.mdx).


OskiTheCoder (approved):
nice

PR #84, opened 2023-04-06
Upgrade Storybook to v7

stephenjwatkins:
- Upgrades Storybook to v7
- Switches configuration files to TypeScript and updates their structure
- Convert MDX to MDX+CSF
- Add AriaProps to button prop types to support `onPress`, etc
- Fix Icon types build issue


easy-ui-icons/vite.config.ts
```diff
@@ -87,15 +87,14 @@ async function getSvgFilenameFromJson(filename: string) {
...
-    }
     declare const ${svgName}: ({
       title,
       titleId,
       ...props
-    }: React.SVGProps<SVGSVGElement> & SVGRProps) => JSX.Element;
+    }: {
+      title?: string;
+      titleId?: string;
+    } & React.SVGProps<SVGSVGElement>) => JSX.Element;
```
stephenjwatkins, line 97: TypeScript has trouble creating downstream definitions when `SVGRProps` is used due to it being an interface internal to this file, so having the structure inline remedies this issue

easy-ui-react/src/Button/Button.mdx
stephenjwatkins, line 1: Storybook essentially generated the MDX + CSF files here with its MDX migrator. we may want to create a separate task at some point to figure out a better Storybook documentation structure overall but this at least replicates our existing pattern in the new v7 format

easy-ui-react/src/utilities/storybook.tsx
```diff
@@ -38,3 +40,27 @@ function getTokensControl(tokenAliases: string[]) {
...
+  <div style={buttonStoryStyles}>
+    <Story />
+  </div>
+);
+
+export const ButtonStoryOnDarkBackgroundDecorator: Decorator = (Story) => (
+  <div style={{ ...buttonStoryOnDarkBackgroundStyles, ...buttonStoryStyles }}>
+    <Story />
+  </div>
+);
```
OskiTheCoder, line 66: **suggestion:** doesn't necessarily have to be done now, can be a follow up, but as you pointed out on a previous PR, we could probably reuse these decorators for other inline components. what are your thoughts on renaming to `inline` and `Inline`?

easy-ui-react/src/IconButton/IconButton.stories.tsx
```diff
@@ -0,0 +1,96 @@
+import AddIcon from "@easypost/easy-ui-icons/Add";
```
OskiTheCoder, line 1: for some reason im unable to view this story via the preview link, do you know why that might be?

Screen Shot 2023-04-11 at 8 41 08 AM

stephenjwatkins: nice catch. had a typo in the story file. even with the experimental MDX language server enabled in VSCode, i'm still finding it lacking with catching errors like that :(

PR #87, opened 2023-04-07
banner spec

OskiTheCoder:
- adds spec for upcoming `Banner` component

### DIFF
--- a/documentation/specs/Banner.md
+++ b/documentation/specs/Banner.md
@@ -1,4 +1,4 @@
-# Banner Component Specification
+# `Banner` Component Specification
 
 ## Overview
 
@@ -6,7 +6,7 @@ The `Banner` component displays a prominent message above the navigation bar at 
 
 ### Use Cases
 
-The primary use case of the `Banner` component is to inform the user of promotions and other offerings on the marketing (logged out) site. **Not** to be used to call attention to the user as a result of interacting with the UI.
+The primary use case of the `Banner` component is to inform the user of promotions and other offerings. **Not** to be used to call attention to the user as a result of interacting with the UI.
 
 ---
 
@@ -15,32 +15,75 @@ The primary use case of the `Banner` component is to inform the user of promotio
 ### API
 
 ```typescript
-export type BannerColor = "primary" | "success" | "neutral";
+export type BannerStatus = "primary" | "success" | "neutral";
 export type BannerProps = {
   /**
-   * Banner color
-   * @default "primary"
+   * Banner status color
+   * @default "success"
    */
-  color?: BannerColor;
+  status?: BannerColor;
   /**
-   * Content to display
+   * Banner title text
+   * @default null
    */
-  children: ReactNode;
+  title?: Text;
+  /**
+   * Banner body text
+   */
+  body: Text;
 };
 ```
 
 ### Example Usage
 
+_Default with title and body_
+
 ```tsx
 import Banner from "@easypost/easy-ui/Banner";
 import Text from "@easypost/easy-ui/Text";
 
-<Banner color="success">
-  <Text variant="subtitle1">
-    Limited Time Only:
-  <Text>
-  <Text variant="body1">
-    Get your first $250 of labels free!
-  <Text>
-</Banner>
+<Banner
+  title={<Text>Limited Time Only:</Text>}
+  body={<Text>Get your first $250 of labels free!</Text>}
+/>;
 ```
+
+_Status with no title_
+
+```tsx
+import Banner from "@easypost/easy-ui/Banner";
+import Text from "@easypost/easy-ui/Text";
+
+<Banner
+  status="primary"
+  body={<Text>No fees on 120,000 packages per year.</Text>}
+/>;
+```
+
+_Status with title_
+
+```tsx
+import Banner from "@easypost/easy-ui/Banner";
+import Text from "@easypost/easy-ui/Text";
+
+<Banner
+  status="neutral"
+  title={<Text>Limited Time Only:</Text>}
+  body={<Text>Get your first $250 of labels free!</Text>}
+/>;
+```
+
+## Behavior
+
+### Accessibility
+
+A `banner` landmark identifies site-oriented content at the beginning of each page within a website. Site-oriented content typically includes things such as the logo or identity of the site sponsor, and site-specific search tool. A banner usually appears at the top of the page and typically spans the full width.
+
+By default, the HTML's `<header>` element has an identical meaning to the banner landmark, unless it is a descendant of `<aside>`, `<article>`, `<main>`, `<nav>`, or `<section>`, at which point `<header>` exposes a generic role, and not the equivalent of the site-wide banner.
+
+Each page may have a `banner` landmark, but each page should generally be limited to a single element with the role of banner. In the case of a page containing nested `document` and/or `application` roles, each nested `document` or `application` role may also have one `banner` landmark. If a page includes more than one `banner` landmark, each should have a unique accessible name.
+
+### Further reading on accessibility
+
+- [W3](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/banner.html)
+- [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/banner_role)
--- a/documentation/specs/Banner.md
+++ b/documentation/specs/Banner.md
@@ -1,0 +1,46 @@
+# Banner Component Specification
+
+## Overview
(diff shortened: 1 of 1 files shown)


documentation/specs/Banner.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Overview
+
+The `Banner` component displays a prominent message above the navigation bar at the top of the page that cannot be dismissed.
+
+### Use Cases
+
+The primary use case of the `Banner` component is to inform the user of promotions and other offerings on the marketing (logged out) site. **Not** to be used to call attention to the user as a result of interacting with the UI.
+
+---
```
OskiTheCoder, line 11: didnt include prior art from polaris as i think our use case does differ a bit, but open to including it and noting the differences

documentation/specs/Banner.md
```diff
@@ -0,0 +1,46 @@
+# Banner Component Specification
+
+## Overview
+
+The `Banner` component displays a prominent message above the navigation bar at the top of the page that cannot be dismissed.
```
OskiTheCoder, line 5: will check again regarding dismissibility to ensure its the behavior we want

documentation/specs/Banner.md
```diff
@@ -0,0 +1,46 @@
...
+
+<Banner color="success">
+  <Text variant="subtitle1">
+    Limited Time Only:
+  <Text>
+  <Text variant="body1">
+    Get your first $250 of labels free!
+  <Text>
+</Banner>
+```
```
OskiTheCoder, line 46: my thinking for this component was that we would use it in conjunction with the `Text` component as opposed to accepting props like boldText, bodyText..etc

documentation/specs/Banner.md
```diff
@@ -0,0 +1,46 @@
...
+### API
+
+```typescript
+export type BannerColor = "primary" | "success" | "neutral";
+export type BannerProps = {
+  /**
+   * Banner color
+   * @default "primary"
+   */
+  color?: BannerColor;
```
stephenjwatkins, line 24: **question:** is there a more semantic name we could use instead of `color`? something that represents more of the state of the banner rather than an effect of the state.

i notice Shopify uses `status` in a similar vein, MUI uses `severity` for its `Alert`, Twilio uses `variant` for its `Callout`, etc

documentation/specs/Banner.md
```diff
@@ -0,0 +1,46 @@
+# Banner Component Specification
+
+## Overview
+
+The `Banner` component displays a prominent message above the navigation bar at the top of the page that cannot be dismissed.
+
+### Use Cases
+
+The primary use case of the `Banner` component is to inform the user of promotions and other offerings on the marketing (logged out) site. **Not** to be used to call attention to the user as a result of interacting with the UI.
```
stephenjwatkins, line 9: **question:** is a banner intended to *only* be used on the marketing site? like exclusively? or does that just happen to be the more common pattern?

i'm wondering how strong of a scope stance we need to take here in the use cases for it

documentation/specs/Banner.md
stephenjwatkins: this is a good question. i like your thought process in general but i feel a little uncertain.

looking at Figma, it almost looks like the "subtitle" is treated as a special thing in the comment. for instance, i notice that on mobile screens it breaks the lines into two, which wouldn't be easy to handle on the consumer side with just two `<Text />` components as the content:

image

it may be worth asking design if the "subtitle" is meant to be an intentional pattern that we should codify in our component API. if it isn't, it would be worth giving thought as to how we would handle the mobile situation where it breaks into two lines

stephenjwatkins (commented):
nice! just a couple questions

documentation/specs/Banner.md
OskiTheCoder: per design, no dismissibility for this version

documentation/specs/Banner.md
OskiTheCoder: good callout, i checked with design and it is indeed an intentional pattern for the component to be able to render semibold and body text in which the semibold text stacks on the body text on mobile devices. for cases when there is no semibold text, the body text should just wrap.

with that said, im curious how do you feel about the updated component API. per design, its a valid use case for either (`title` and `body`) to be able to handle a text link. otherwise i would've just made `title` a string and deferred the `body` to `children`. of course, we can still do something similar,

### API

```typescript
export type BannerStatus = "primary" | "success" | "neutral";
export type BannerProps = {
  /**
   * Banner status color
   * @default "success"
   */
  status?: BannerColor;
  /**
   * Banner title text
   * @default null
   */
  title?: Text;
  /**
   * Banner body content
   */
  children: ReactNode;
};
```
### Usage

```tsx
import Banner from "@easypost/easy-ui/Banner";
import Text from "@easypost/easy-ui/Text";

<Banner title={<Text>Limited Time Only:</Text>}>
Get your first $250 of labels free!
</Banner>
;
```

maybe thats a bit better, curious how you feel about it though.

documentation/specs/Banner.md
OskiTheCoder: i went ahead and changed it to `status`

documentation/specs/Banner.md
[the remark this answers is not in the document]
OskiTheCoder: great point, included a section on `Accessibility`

documentation/specs/Banner.md
OskiTheCoder: to my understanding, its scope is limited to the marketing site, but on second thought, i do think perhaps its not really necessary for us to take that sort of stance for the docs

documentation/specs/Banner.md
stephenjwatkins, line 1: **question:** do you think we need to take into account the space around the banner at all?

image

i notice there are specs for that aspect in Figma:

image

i realize it's spacing around a `<Banner />` and not part of the inners of the actual component but i also am a little wary of relying on outside consumers sticking to the specs defined here. i wonder if it'd be worth exposing some kind of container component as well that adheres to this spec? `<Banner.Container />` or some such. ehh probably not, probably that's a responsibility of a layout component outside of the Banner but i'll mention it here in case...

documentation/specs/Banner.md
OskiTheCoder: sounds good, updated spec to reflect pattern with using `children`. i also updated the previous `title` prop to `emphasis`. i feel the same about coming up with a name for that one lol, but i think `emphasis` is a bit more contextually appropriate.

documentation/specs/Banner.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Behavior
+
+### Accessibility
+
+A `banner` landmark identifies site-oriented content at the beginning of each page within a website. Site-oriented content typically includes things such as the logo or identity of the site sponsor, and site-specific search tool. A banner usually appears at the top of the page and typically spans the full width.
+
+By default, the HTML's `<header>` element has an identical meaning to the banner landmark, unless it is a descendant of `<aside>`, `<article>`, `<main>`, `<nav>`, or `<section>`, at which point `<header>` exposes a generic role, and not the equivalent of the site-wide banner.
+
+Each page may have a `banner` landmark, but each page should generally be limited to a single element with the role of banner. In the case of a page containing nested `document` and/or `application` roles, each nested `document` or `application` role may also have one `banner` landmark. If a page includes more than one `banner` landmark, each should have a unique accessible name.
```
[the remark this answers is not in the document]
OskiTheCoder: updated to reflect that the `banner` landmark is not specific to this component

documentation/specs/Banner.md
OskiTheCoder: its something i thought about as i was writing this spec and as we go to the implementation phase. my thinking was that the `Banner` component wouldn't handle its surrounding space and we'd have some layout container to expose to handle that as you alluded to

documentation/specs/Banner.md
```diff
@@ -0,0 +1,75 @@
...
+The primary use case of the `Banner` component is to inform the user of new features, promotions, and other offerings. **Not** to be used to call attention to the user as a result of interacting with the UI.
+
+---
+
+## Design
+
+### API
+
+```typescript
+export type BannerStatus = "primary" | "success" | "neutral";
```
OskiTheCoder, line 18: **thought** i have primary here, but wondering if promotional would make more sense, not sure

stephenjwatkins: i think what you have here makes sense because it matches the states we have for `Button`s. i think that consistency will be important for consumers

image

stephenjwatkins (approved):
nice! feel like we got to a solid place with the API

PR #93, opened 2023-04-10
Create Tooltip spec

stephenjwatkins:


documentation/specs/Tooltip.md
```diff
@@ -0,0 +1,248 @@
...
+- Custom placement
+- Ability to disable independent of its trigger
+- Custom trigger mode (hover and focus, or just focus)
+- Ability to be controlled or uncontrolled
+
+### Risks and Challenges
+
+- Ensuring tooltips don't get cut off within hidden containers
+- Keeping the API elegantly simple while needing to be aware of inner element `ref`s and needing to pass `prop`s to inner components
+- Will need to ensure that Easy UI components are forwarding refs in order to handle ref attachment
```
stephenjwatkins, line 22: this will require forwarding refs in our `*Button` components

documentation/specs/Tooltip.md
```diff
@@ -0,0 +1,248 @@
...
+  onOpenChange?: (isOpen: boolean) => void;
+
+  /**
+   * The placement of the element with respect to its anchor element.
+   * @default 'top'
+   */
+  placement?: Placement;
+
+  /** By default, opens for both focus and hover. Can be made to open only for focus. */
+  trigger?: "focus";
```
OskiTheCoder, line 70: **question** is this intended to be optional? if so, would it be worth typing?

documentation/specs/Tooltip.md
```diff
@@ -0,0 +1,248 @@
...
+};
+```
+
+### Anatomy
```
OskiTheCoder, line 74: really like the inclusion of anatomy for this component

documentation/specs/Tooltip.md
```diff
@@ -0,0 +1,248 @@
...
+### API
+
+React Aria exposes a lot of functionality for placement and behavior, but the API for Easy UI's `Tooltip` component will be intentionally limited to reduce inconsistency across the UIs. Instead, the component will hold the specific React Aria configuration optimal for EasyPost. We can expose functionality on a per-needed basis as it arises in the future.
+
+```ts
+// prettier-ignore
+type Placement = 'bottom' | 'bottom left' | 'bottom right' | 'bottom start' | 'bottom end' |
+    'top' | 'top left' | 'top right' | 'top start' | 'top end' |
+    'left' | 'left top' | 'left bottom' | 'start' | 'start top' | 'start bottom' |
+    'right' | 'right top' | 'right bottom' | 'end' | 'end top' | 'end bottom';
```
OskiTheCoder, line 42: **question**:  im trying to visualize these placements, with respect to `bottom` does this diagram align with whats here?

Screen Shot 2023-04-13 at 1 08 58 PM

OskiTheCoder (approved):
just one additional question, but its non blocking, great work!

documentation/specs/Tooltip.md
stephenjwatkins: good question. again it is a copy from the same type that's used on Aria's side: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-types/overlays/src/index.d.ts#L16

`bottom` by itself is center like you drew. the difference between `start`, `end`, `left`, and `right` are which values respect the user's language direction. as an example, if you want it to be anchored to the "start" of an element dependent on LTR or RTL support, you'd use `start`; whereas if you wanted it to always be on the "left", you'd use `left`

PR #103, opened 2023-04-12
Create Menu spec

stephenjwatkins:


documentation/specs/PopupMenu.md
stephenjwatkins, line 1: there were a couple interesting findings from this:

(a) i didn't see the name `PopupMenu` in my research for this kind of component. i saw `Menu` and `ActionMenu`, which makes me wonder if we should follow suit and name it one of those names instead. i don't feel strongly about it so curious if there are any strong leanings either way from anyone else

(b) i noticed that there are different kinds of wiring and accessibility considerations for a Menu of options that gets triggered from a button like in this component's case versus a menu that's linked to an input like a dropdown. as such, most component libraries have different components to represent these two different interactions. React Aria for instance uses a `useComboBox` set of primitives to build a dropdown with linked menu of options and uses `useMenu` for a standard list of menu items. i mention this because in Figma they appear to be treated the same, but in reality they will be wired differently. however care should be put in place to be able to reuse what elements we can across the different wirings

ldewald: +1 for `Menu` or `ActionMenu`

OskiTheCoder: slight preference for `Menu`

documentation/specs/Menu.md
```diff
@@ -0,0 +1,391 @@
...
+        <Menu.Section>
+          <Menu.Item key="edit">Edit</Menu.Item>
+          <Menu.Item key="duplicate">Duplicate</Menu.Item>
+        </Menu.Section>
+        <Menu.Section>
+          <Menu.Item key="copy">Copy</Menu.Item>
+          <Menu.Item key="cut">Cut</Menu.Item>
+          <Menu.Item key="paste">Paste</Menu.Item>
+        </Menu.Section>
+      </Menu.Menu>
```
stephenjwatkins, line 140: i updated the component to `Menu` which i like. however, i'm not sure i love `<Menu.Menu>` to signify the part of the component that holds the items just because it looks a little strange to repeat the word. other options could be `<Menu.Items>` or `<Menu.Overlay>` or something else?

OskiTheCoder: hmm, id be ok with `<Menu.Items>` as it defines the collection/element relationship with `<Menu.Item>` but also understand that its not ideal either.  i know in an earlier comment I said I prefer `Menu` to `ActionMenu` but id also be fine if you changed it up to `ActionMenu` to avoid the repeated word.

stephenjwatkins: checking other libraries, i see these names used for the same component:

`<Menu.List>`, `<Menu.Overlay>`, and `<Menu.Menu>`

any stand out to you / make more sense than the others?

OskiTheCoder: I think I have a slight preference for `<Menu.List>` over the others, but i also dont particularly love it either.

stephenjwatkins:
> everything looks good to me and it fits what we need as far as i can tell. admittedly i think some of the `react-aria` dependencies feel a little heavy, but i think thats almost to be expected for a more intricate component

i agree that it feels like a lot. if you look at the examples you'll see that there is in fact a lot of wiring required unfortunately. i'm glad it offers flexibility but yeah it is a lot. React Aria Components will probably help with this once they're out of beta

PR #108, opened 2023-04-13
Use Sass @use instead of @import

ldewald:


stephenjwatkins (commented):
nice! looks good overall. just one question:

i don't have a super strong opinion about this but i'm wondering if for our "common" helpers in our `.module.scss` files if we should `@use "../styles/common" as *`. i don't think the `c.design-token()`, etc is too bad, but i wonder if it may make sense to directly expose the helpers forwarded in common to make it slightly easier to use 🤔 curious what you / others think?

ldewald:
> nice! looks good overall. just one question:
> 
> i don't have a super strong opinion about this but i'm wondering if for our "common" helpers in our `.module.scss` files if we should `@use "../styles/common" as *`. i don't think the `c.design-token()`, etc is too bad, but i wonder if it may make sense to directly expose the helpers forwarded in common to make it slightly easier to use 🤔 curious what you / others think?

I wondered the same thing and went back and forth myself so updated with this pattern instead

easy-ui-react/src/Button/Button.module.scss
```diff
@@ -1,5 +1,5 @@
-@use "./mixins" as Button;
-@import "../styles/common";
+@use "../styles/common" as *;
```
ldewald, line 1: Swapped the order because it felt better to have it come before the namespaced `@use`

stephenjwatkins: agreed!

stephenjwatkins (approved):
nice!

PR #117, opened 2023-04-17
notification spec

OskiTheCoder:
`Notification` component spec


documentation/specs/Toast.md
```diff
@@ -0,0 +1,174 @@
...
+The `Toast` component displays a brief non-disruptive message to the user as a result of an action taken. The displayed message is accompanied by an associated `status` icon and cannot be manually dismissed. **Not** to be used to display critical information such as a system failure, see the `Alert` component for such behavior.
+
+### Use Cases
+
+- Inform user of a completed action such as an account creation.
+- Inform user that an asynchronous process has been triggered based on their action.
+
+### Features
+
+- ARIA `role="status"`
```
OskiTheCoder, line 14: react-aria has a `useToast` hook, but seems like its not the aria pattern we want for our toast and might be tricky as to where we draw the line for whats within this library vs what is left up to the consumer https://react-spectrum.adobe.com/react-aria/useToast.html

documentation/specs/Toast.md
```diff
@@ -0,0 +1,174 @@
...
+   * Toast status color
+   * @default "success"
+   */
+  status?: ToastStatus;
+  /**
+   * Toast content
+   */
+  children?: ReactNode;
+};
+```
```
OskiTheCoder, line 45: i know some libraries expose a `duration` property for the `Toast` and im open to adding it, but im having trouble envisioning a scenario where we add the property and have minimal state configuration within `Toast`

stephenjwatkins: i think we'll want to try to encapsulate the duration config within Easy UI, at least until we have a concrete use case for supporting multiple durations outside of the library

documentation/specs/Toast.md
```diff
@@ -0,0 +1,174 @@
...
+- Building for the shipper's experience at EasyPost according to internal specs while maintaining a pleasant DX for outside consumers.
+
+---
+
+## Design
+
+### API
+
+```typescript
+export type ToastStatus =
```
stephenjwatkins, line 28: i'm second guessing the naming of the status variants now that i see them on this component. i know i mentioned in the `Banner` component that it's nice that they align with the `Button` states for consistency, but i'm wondering if these would make more sense named how they are in Figma.

what do you think? do you think developers will understand what these status variants mean as-is or more-so as they are outlined in Figma?

documentation/specs/Toast.md
```diff
@@ -0,0 +1,174 @@
...
+
+_Default_
+
+```tsx
+import Toast from "@easypost/easy-ui/Toast";
+
+<Toast>Account Created</Toast>;
+```
+
+As mentioned above, it is unlikely the `Toast` component will be used in this way. Below is a relatively detailed example for how one might use the `Toast` component. Note that this is **not** production ready, nor is it the best solution, but it is a viable starting point.
```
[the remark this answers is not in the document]
stephenjwatkins: i think we should handle the state management for the toasts in Easy UI that way implementation doesn't have to be up to consumers where it's easy to get accessibility and other behaviors wrong. if we did that, i don't think the examples would be as relevant

documentation/specs/Toast.md
stephenjwatkins: i do think we will want to utilize `useToast` both here and for the `Alert` component with some minor configuration differences between the two. it handles a lot of things that we will want to handle in Easy UI, such as rendering it into a labeled region, automatically focusing in and out, attaching accessibility labels, state management, and timers.

from what i understand, the aspect that we would need to adjust for in our `Toast` component would be setting the `role="status"` on the toast element directly, overriding the `alert` value that [comes from Aria](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/toast/src/useToast.ts#L91). from what i can tell, the rest of it would be the same.

let me know if i'm misunderstanding anything related to the accessibility/Aria situation.

documentation/specs/Toast.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+## Behavior
+
+### Accessibility
+
+The `Toast` component has an ARIA `role="status"` and elements with the role status have an implicit `aria-live` value of `polite` and an implicit `aria-atomic` value of `true`. Toasts are **not** suitable for dynamically changing content, consider using the `AlertBanner` component for such a use case.
+
+The `Toast` component does not provide a way for the user to manually dismiss the message, hence toasts should timeout after some time. When making decisions on the timeout duration, consider that users have different reading speeds, vision levels, and literacy levels. Avoid messages longer than a sentence and avoid rendering interactive elements within toasts that require the user to act.
```
stephenjwatkins, line 174: i noticed some libraries have special messages relating to toasts that show errors. like discouraging automatic dismissal or increasing the automatic dismissal duration. out of curiosity, do we need any special consideration for error toasts or is it the same as other toasts?

documentation/specs/Toast.md
OskiTheCoder: nope, youre good! i updated to use `react-aria` and `react-stately`. my initial hesitation for using these was for our use of the `Toast`, i wasn't sure if keyboard navigation was a necessity or would cause harm seeing as we wouldn't want to render anything interactive for an element that per our spec, needs to time out; but i think i had a misunderstanding initially about the purpose of the region landmark. for `react-stately`, it initially looked like we'd have to do some heavy configuration, but it ended up being not too bad.

documentation/specs/Toast.md
OskiTheCoder: thats a good point, how do you feel about `variant`  and the renaming? 

my only worry is that `neutral` is an available property value to control colors in `Button` and maps to something entirely different

OskiTheCoder: on second thought, i changed it back to roughly align with the `Button` states, im open to a different naming or convention, but i realized that for the `Banner` the `blue-800` variation seems ideal to be named `neutral` and i feel like it might be confusing to have `neutral` as purple for `Alert` and `Toast` but have `neutral` as `blue-800` for the `Banner`.  open to ideas on this though if you have an alternative representation in mind

documentation/specs/Toast.md
```diff
@@ -0,0 +1,321 @@
...
+  message: string;
+};
+
+export type ToastStateProps = AriaToastProps<ToastProps> & {
+  toast: QueuedToast<ToastProps>;
+  state: ToastState<ToastProps>;
+};
+```
+
+### Anatomy
```
OskiTheCoder, line 67: anatomy is missing some details like name destructuring from props, passing props to `Text`, and passing `symbol` to `Icon`. let me know if you'd like to see them added now as opposed to returning to add them later

documentation/specs/Toast.md
stephenjwatkins, line 1: - not a huge deal but do you think we need to leverage the global toast queue from Aria? i was thinking that we could get away with using a single provider at the root of `EasyUIProvider` and create a lightweight hook around their `useToastState()` hook, keeping the toast invocations within the React tree and not a separate export that can be used anywhere

- do we feel confident that consumers will use the `<ToastContainer />` correctly? will it need to render into a portal or use an `OverlayContainer` from Aria? how will we ensure that it's positioned properly in the App and renders over the elements correctly?

- how separate are we wanting toasts and alerts in EasyUI? for example, will consumers need to do the setup for both toasts and alerts independently? `<ToastContainer />`, `<AlertContainer />`, `ToastQueue`, `AlertQueue`, etc. seems like it could become unfriendly

it makes me think that we need a more generic `Notification` system that can issue toasts or alerts. thinking out loud:

```tsx
function App() {
  return (
    <>
      <Nav />
      <Notifications />
      <Main />
      <Footer />
    </>
  );
}
```

```tsx
function Component() {
  const notifications = useNotifications();
  const onToast = () => {
    notifications.toast("Error");
  };
  const onAlert = () => {
    notifications.alert("Password reset sent!");
  };
  return (<div></div>);
}
```

each invocation would go to the same Notifications queue and Notifications region.

OskiTheCoder: updated spec to reflect these ideas. specifically, instead of seperate components and containers, we now have a generic `Notification` component, rendering of `NotificationProvider` will now happen at the root of our provider, `EasyUIProvider` and toast/alert invocations happen via a`useNotification`hook

documentation/specs/Notification.md
```diff
@@ -0,0 +1,567 @@
...
+import type { QueuedToast } from "@react-stately/toast";
+
+export type NotificationType = "alert" | "toast";
+
+export type NotificationStatus =
+  | "promotional"
+  | "success"
+  | "neutral"
+  | "error"
+  | "warning";
```
OskiTheCoder, line 56: i hesitated a little to use these names since `neutral` is a property on the `Button` that means something entirely different, but based on this architecture, consumers shouldn't really have to worry about passing these props directly and instead they'll use `showNeutralToast` and `showNeutralAlert`, which i think is a bit better, but open to ideas.

documentation/specs/Notification.md
```diff
@@ -0,0 +1,567 @@
...
+   * interact with the queue object.
+   */
+  state: NotificationState<NotificationProps>;
+};
+
+/** The methods below are accessible on the object returned by the `useNotification` hook */
+export type NotificationState<NotificationProps> =
+  ToastState<NotificationProps> & {
+    /** Shows promotional color status toast */
+    showPromotionalToast(content: NotificationProps): void;
```
stephenjwatkins, line 94: **thought:** since each method defines both the `type` and the `status`, i guess we would want a type that doesn't include those properties to be used on `content` instead. something like the below.

```
type NotificationContent = Omit<NotificationProps, "type" | "status">;

showPromotionalToast(content: NotificationContent)
```

documentation/specs/Notification.md
```diff
@@ -0,0 +1,567 @@
...
+  top?: string;
+  right?: string;
+  bottom?: string;
+  left?: string;
+};
+```
+
+### Anatomy
+
+The `Notification` component's functionality will be handled by React Aria's `useToast` hook, which will assist in managing behavior and accessibility. Individual notifications will be rendered in a landmark region and handled by a `NotificationRegion` component, which behind the scenes is leveraging React Aria's `useToastRegion` hook. The `NotificationContainer` component handles rendering of the `NotificationRegion` component. Queue management will be handled by `EasyUINotificationQueue`, which extends React Stately's `ToastQueue`. The queue is bundled into a `useNotificationState` hook which is using React Stately's `useToastQueue` hook. A `NotificationProvider` consumes the `useNotificationState` hook and provides the returned value to its children via React's `useContext` hook. This will be rendered at the root of `EasyUIProvider`. The `useContext` hook is not directly exposed to consumers; it is wrapped up in a `useNotification` hook on which the returned object can invoke functions to display toasts or alerts in a `NotificationContainer`.
```
stephenjwatkins, line 138: quite the wiring 😂 😰

documentation/specs/Notification.md
```diff
@@ -0,0 +1,567 @@
...
+          )
+        : null}
+    </>
+  );
+}
+```
+
+### Example Usage
+
+The examples below assume your app is consuming `EasyUIProvider`. This is important because `NotificationProvider` is rendered at the root of `EasyUIProvider`.
```
stephenjwatkins, line 360: **thought:** fwiw the Provider will be required for a lot of different functionality. we'll make it clear in the installation instructions that it's a required piece of integrating Easy UI.

documentation/specs/Notification.md
```diff
@@ -0,0 +1,567 @@
...
+  /**
+   * Notification status
+   * @default "success"
+   */
+  status?: NotificationStatus;
+  /**
+   * Notification will render with left aligned status icon
+   * @default false
+   */
+  noIcon?: boolean;
```
stephenjwatkins, line 74: **nit:** not a huge deal but i'll just mention that i'm slightly on the fence with this name. i wonder if `hasIcon` and defaulting the opposite way would be better, but not entirely sure. we can see how it feels during implementation and if anyone else has any other thoughts.

documentation/specs/Notification.md
```diff
@@ -0,0 +1,567 @@
...
+  | "error"
+  | "warning";
+
+export type NotificationProps = {
+  /** Notification type */
+  type: NotificationStatus;
+  /** Notification message */
+  message: ReactNode;
+  /** Callback function when dismissing alerts */
+  onDismiss?: () => void;
```
stephenjwatkins, line 64: **suggestion:** we should have a separate `AlertType` and `ToastType` that extend `NotificationType`, and ensure that properties only relevant to each are included. `onDismiss` for instance should only be on alerts. then we can ensure that we pass in `AlertType` as the `content` type for the `show*Alert()` methods and likewise with the toast methods

documentation/specs/Notification.md
```diff
@@ -0,0 +1,567 @@
...
+export type NotificationType = "alert" | "toast";
+
+export type NotificationStatus =
+  | "promotional"
+  | "success"
+  | "neutral"
+  | "error"
+  | "warning";
+
+export type NotificationProps = {
```
stephenjwatkins, line 58: **question:** do toasts need to accommodate custom timing by the consumers or will that all be encapsulated by our library? i think if it can be encapsulated by us that would be better but just wanted to double check.

documentation/specs/Notification.md
OskiTheCoder: good idea, i think this is something ill do when it comes time for implementation

documentation/specs/Notification.md
OskiTheCoder: agree, and it will be encapsulated by our library automatically when invoking toasts. 

react stately's `ToastQueue` has an `add` method that accepts `ToastOptions` as an optional argument and for us, when toasts are invoked, we'll pass along the options for timing when `add` is called behind the scenes.

OskiTheCoder:
i agree with pretty much all of the suggestions and will include it during the implementation phase and update the spec as needed at that stage.

PR #141, opened 2023-04-25
decision record for unstyled button

OskiTheCoder:
- adds decision record for supporting an internal `UnstyledButton` component

### DIFF
--- a/documentation/decisions/011_unstyled_button.md
+++ b/documentation/decisions/011_unstyled_button.md
@@ -35,39 +35,47 @@ Similar to the `Button`, `DropdownButton`, and `IconButton` components, the `Uns
 
 ```ts
 export type UnstyledButtonProps = AriaButtonProps<"button"> & {
-  /** Content to render */
-  children?: ReactNode;
   /* Classname to apply styles to button */
   className?: string;
-  /**
-   * Disables button
-   * @default false
-   */
-  isDisabled?: boolean;
+  /** Link's destination */
+  href?: string;
 };
 ```
 
 ### Anatomy
 
 ```tsx
-import { useButton } from "react-aria";
+import { forwardRef } from "react";
+import { mergeRefs } from "@react-aria/utils";
+import { useButton, mergeProps } from "react-aria";
 import { classNames } from "../utilities/css";
+import { omitReactAriaSpecificProps } from "./utilities";
 
-export function UnstyledButton(props: UnstyledButtonProps) {
-  const { className, isDisabled = false, children = "Button" } = props;
+export const UnstyledButton = forwardRef<null, UnstyledButtonProps>((props, inRef)) => {
+  const {
+    className,
+    href = "",
+    isDisabled = false,
+    children = "Button",
+    ...restProps
+  } = props;
 
   const ref = useRef(null);
-  const { buttonProps } = useButton(props, ref);
+  const As = href ? "a" : "button";
+  const { buttonProps: elementProps } = useButton(
+    { ...props, elementType: As },
+    ref,
+  );
 
   return (
-    <button
+    <As
       disabled={isDisabled}
-      ref={ref}
+      ref={mergeRefs(ref, inRef)}
       className={classNames(className)}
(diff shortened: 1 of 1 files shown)


documentation/decisions/011_unstyled_button.md
```diff
@@ -0,0 +1,99 @@
...
+  /** Content to render */
+  children?: ReactNode;
+  /* Classname to apply styles to button */
+  className?: string;
+  /**
+   * Disables button
+   * @default false
+   */
+  isDisabled?: boolean;
+};
```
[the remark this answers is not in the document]
stephenjwatkins: **thought:** my thinking was that it would carry the logic that we have in `<Button />` for rendering a `button` or `a` element based on the `href`, then we could use this in `<Button />` and `<IconButton />` to keep the button logic DRY, and if we needed that logic internally for other components we would already have that behavior complete.

admittedly you know more about the internals of our buttons, so deferring to you.

documentation/decisions/011_unstyled_button.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+### Anatomy
+
+```tsx
+import { useButton } from "react-aria";
+import { classNames } from "../utilities/css";
+
+export function UnstyledButton(props: UnstyledButtonProps) {
```
stephenjwatkins, line 56: **thought:** we'll want to make sure it forwards refs properly. You can see my `<Tooltip />` pull request for how that can be achieved with Aria

documentation/decisions/011_unstyled_button.md
```diff
@@ -0,0 +1,99 @@
...
+  const ref = useRef(null);
+  const { buttonProps } = useButton(props, ref);
+
+  return (
+    <button
+      disabled={isDisabled}
+      ref={ref}
+      className={classNames(className)}
+      {...buttonProps}
+    >
```
stephenjwatkins, line 68: **thought:** we'll want to ensure that all passed in props are spread on the component. i found that with Aria the `buttonProps` returned are only the ones relevant to the Aria library, and it removes some props that are still important to being passed into the element. if you see my `<Tooltip />` pull request, you'll notice that I spread the `...restProps` and `mergeProps()` onto the underlying element to ensure the appropriate props are passed in.

stephenjwatkins (approved):
seems conceptually sound

documentation/decisions/011_unstyled_button.md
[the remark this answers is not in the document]
ldewald: > **thought:** my thinking was that it would carry the logic that we have in `<Button />` for rendering a `button` or `a` element based on the `href`, then we could use this in `<Button />` and `<IconButton />` to keep the button logic DRY, and if we needed that logic internally for other components we would already have that behavior complete.

+1 this would be great to include

documentation/decisions/011_unstyled_button.md
stephenjwatkins: this problem is slightly tricky. i ended up needing to make a helper in `Button/utilities` as part of the Tooltip PR to remove React Aria-specific props before spreading onto the `<button />` element. i'm open to better approaches.

i mentioned it in the Tooltip PR, but the problem is that the Aria `useButton()` hook swallows some props that need to get passed to the `<button />` element. most notably, the `onPointerEvent` and `onPointerLeave` props provided by the Tooltip trigger handler. essentially that means we have to ensure those props get spread to `<button />` independent of the `buttonProps` returned by the `useButton()` hook.

documentation/decisions/011_unstyled_button.md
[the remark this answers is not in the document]
OskiTheCoder: agree that would be a solid approach

documentation/decisions/011_unstyled_button.md
OskiTheCoder: interesting, your comment makes sense and i went ahead and included the utility function. ill take a closer look on the `Tooltip` PR

PR #146, opened 2023-04-26
Create tooltip

stephenjwatkins:
Creates `<Tooltip />`.

- [x] Code complete
- [x] Unit tests
- [x] Component TSDoc written
- [x] Storybook stories and docs
- [x] Specs and documentation up-to-date
- [x] Add changeset

Other:

- Creates storybook `OverlayLayoutDecorator` for stories with overlays
- Forwards ref on button components for attaching overlays
- Fixes storybook breakpoints parsing
- Fixes incompatibility with vitest timers


OskiTheCoder:
haven't looked at the PR yet, but does this need a `Minor` changeset or not since its in pre-release?

stephenjwatkins:
> haven't looked at the PR yet, but does this need a `Minor` changeset or not since its in pre-release?

it does. thanks for catching that. i'm going to make a PR template so i don't keep forgetting things

easy-ui-react/src/Button/utilities.ts
```diff
@@ -17,3 +18,16 @@ export function logWarningIfInvalidColorVariantCombination(
     );
   }
 }
+
+/**
+ * Removes aria-specific props from being applied to button DOM element.
+ */
+export function omitReactAriaSpecificProps(props: object) {
```
stephenjwatkins, line 25: i don't love this. let me know if you have any better ideas.

the problem is that `useButton(props)` is swallowing some props that need to get passed into the `<button />` element. specifically the `onPointerEnter` and `onPointerLeave` events that are being supplied by the Tooltip. they get passed into `useButton()` but don't come out the other side.

ldewald:
**Question:** What is the expected behavior for hiding/showing when going from hovering on one trigger to another trigger? 

While poking at the Storybook, it seems like the initial mouseover has a delay before displaying the tooltip then subsequent mouseovers display the tooltip immediately. 
tooltip-placement

stephenjwatkins:
> While poking at the Storybook, it seems like the initial mouseover has a delay before displaying the tooltip then subsequent mouseovers display the tooltip immediately.

i was going to note that in my review comment. it has a warmup and cooldown period. it's noted in the Spectrum guidelines here: https://spectrum.adobe.com/page/tooltip/#Warmup-and-cooldown

it's using the default from Aria currently, which is 1.5s on enter and .5s on close, but we can customize that to our own liking. i was going to ask Rachel her thoughts

easy-ui-react/src/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,74 @@
...
+  @include component-token("tooltip", "shadow", theme-token("shadow.overlay"));
+  @include component-token(
+    "tooltip",
+    "background",
+    theme-token("color.surface.overlay")
+  );
+  @include component-token(
+    "tooltip",
+    "border_radius",
+    design-token("shape.border_radius.base")
```
OskiTheCoder, line 15: **question**: out of curiosity, is there a benefit to wrapping design tokens in component tokens?

ldewald (approved):
Nice

easy-ui-react/src/Button/utilities.ts
OskiTheCoder: i think what you did here is sensible

easy-ui-react/src/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,74 @@
...
+  );
+  @include component-token("tooltip", "color", theme-token("color.text"));
+
+  background: component-token("tooltip", "background");
+  border-radius: component-token("tooltip", "border_radius");
```
OskiTheCoder, line 20: looks like our design specs need to be updated, ive noticed the `border_radius` discrepancy in a few places

easy-ui-react/src/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,74 @@
+@use "../styles/common" as *;
+
+.Tooltip {
+  @include component-token("tooltip", "arrow-size", "1rem");
+  @include component-token("tooltip", "max-width", 320px);
```
OskiTheCoder, line 5: **question** is this `320px` spec'd somewhere?

OskiTheCoder (approved):
great work! just one non-blocking question

OskiTheCoder:
this is probably more or less just a storybook thing, but i noticed when i zoomed in and out, the tooltip moved to the left. can't seem to replicate this everytime though, so my assumption is that its more or less just a storybook issue, but thought id mention it just in case 

https://user-images.githubusercontent.com/44451175/235483399-a480ada4-06c4-4f1d-9c8b-644df28722d8.mov

easy-ui-react/src/Tooltip/Tooltip.module.scss
stephenjwatkins: good catch. it's not. i wanted to guard against tooltip overflow on mobile screens, but i figured out a better way to handle that. this is now removed

easy-ui-react/src/Tooltip/Tooltip.module.scss
stephenjwatkins: ah good call

stephenjwatkins:
> this is probably more or less just a storybook thing, but i noticed when i zoomed in and out, the tooltip moved to the left. can't seem to replicate this everytime though, so my assumption is that its more or less just a storybook issue, but thought id mention it just in case

good catch. this appears to be an Aria bug:

https://github.com/adobe/react-spectrum/issues/3678
https://github.com/adobe/react-spectrum/issues/2700

fortunately it's likely not a showstopper but something we should keep an eye on

PR #155, opened 2023-04-28
create banner

OskiTheCoder:
Creates `Banner` component 

- [x] Code complete
- [x] Unit tests
- [x] Component TSDoc written
- [x] Storybook stories and docs
- [x] Documentation up-to-date
- [x] Add changeset

### DIFF
--- a/documentation/specs/Banner.md
+++ b/documentation/specs/Banner.md
@@ -15,13 +15,13 @@ The primary use case of the `Banner` component is to inform the user of new feat
 ### API
 
 ```typescript
-export type BannerStatus = "primary" | "success" | "neutral";
+export type BannerVariant = "primary" | "success" | "neutral";
 export type BannerProps = {
   /**
-   * Banner status color
+   * Banner variant color
    * @default "success"
    */
-  status?: BannerStatus;
+  variant?: BannerVariant;
   /** Banner emphasis text */
   emphasis?: ReactNode;
   /** Banner content */
@@ -37,27 +37,27 @@ _Default with emphasis and body_
 import Banner from "@easypost/easy-ui/Banner";
 import Text from "@easypost/easy-ui/Text";
 
-<Banner emphasis={<>Limited Time Only:</>}>
+<Banner emphasis={<>Limited Time Only</>}>
   Get your first $250 of labels free!
 </Banner>;
 ```
 
-_Status with no emphasis_
+_Primary variant with no emphasis_
 
 ```tsx
 import Banner from "@easypost/easy-ui/Banner";
 import Text from "@easypost/easy-ui/Text";
 
-<Banner status="primary">No fees on 120,000 packages per year.</Banner>;
+<Banner variant="primary">No fees on 120,000 packages per year.</Banner>;
 ```
 
-_Status with emphasis_
+_Neutral variant with emphasis_
 
 ```tsx
 import Banner from "@easypost/easy-ui/Banner";
 import Text from "@easypost/easy-ui/Text";
 
-<Banner status="neutral" emphasis={<>Limited Time Only:</>}>
+<Banner variant="neutral" emphasis={<>Limited Time Only</>}>
   Get your first $250 of labels free!
 </Banner>;
 ```
--- a/documentation/specs/Banner.md
+++ b/documentation/specs/Banner.md
@@ -15,13 +15,13 @@ The primary use case of the `Banner` component is to inform the user of new feat
 ### API
 
 ```typescript
-export type BannerVariant = "primary" | "success" | "neutral";
+export type BannerColor = "primary" | "success" | "neutral";
 export type BannerProps = {
   /**
-   * Banner variant color
+   * Banner color color
    * @default "success"
    */
-  variant?: BannerVariant;
+  color?: BannerColor;
   /** Banner emphasis text */
   emphasis?: ReactNode;
   /** Banner content */
@@ -42,22 +42,22 @@ import Text from "@easypost/easy-ui/Text";
 </Banner>;
 ```
 
-_Primary variant with no emphasis_
+_Primary color with no emphasis_
 
 ```tsx
 import Banner from "@easypost/easy-ui/Banner";
 import Text from "@easypost/easy-ui/Text";
 
-<Banner variant="primary">No fees on 120,000 packages per year.</Banner>;
+<Banner color="primary">No fees on 120,000 packages per year.</Banner>;
 ```
 
-_Neutral variant with emphasis_
+_Neutral color with emphasis_
 
 ```tsx
 import Banner from "@easypost/easy-ui/Banner";
 import Text from "@easypost/easy-ui/Text";
 
-<Banner variant="neutral" emphasis={<>Limited Time Only</>}>
+<Banner color="neutral" emphasis={<>Limited Time Only</>}>
   Get your first $250 of labels free!
 </Banner>;
 ```
--- a/documentation/specs/Banner.md
+++ b/documentation/specs/Banner.md
@@ -18,7 +18,7 @@ The primary use case of the `Banner` component is to inform the user of new feat
 export type BannerColor = "primary" | "success" | "neutral";
 export type BannerProps = {
   /**
-   * Banner color color
+   * Banner color
    * @default "success"
    */
   color?: BannerColor;
--- a/documentation/specs/Banner.md
+++ b/documentation/specs/Banner.md
@@ -22,13 +22,9 @@ export type BannerProps = {
    * @default "success"
    */
   status?: BannerStatus;
-  /**
-   * Banner emphasis text
-   */
+  /** Banner emphasis text */
   emphasis?: ReactNode;
-  /**
-   * Banner content
-   */
+  /** Banner content */
   children?: ReactNode;
 };
 ```
--- a/easy-ui-react/src/Banner/Banner.module.scss
+++ b/easy-ui-react/src/Banner/Banner.module.scss
@@ -8,7 +8,7 @@
   );
   @include component-token(
     "banner",
-    "background.color",
+    "color.background",
     theme-token("color.background.success")
   );
 }
@@ -25,7 +25,7 @@
 .variantPrimary {
   @include component-token(
     "banner",
-    "background.color",
+    "color.background",
     theme-token("color.background.primary")
   );
 }
@@ -33,7 +33,7 @@
 .variantNeutral {
   @include component-token(
     "banner",
-    "background.color",
+    "color.background",
     theme-token("color.background.neutral")
   );
 }
@@ -44,6 +44,7 @@
     "border_radius",
     design-token("shape.border_radius.base")
   );
+  @include component-token("banner", "max-width", "1348px");
   display: flex;
   flex-direction: column;
   align-items: center;
@@ -52,9 +53,10 @@
   flex-wrap: wrap;
   white-space: pre-wrap;
   width: 100%;
+  max-width: component-token("banner", "max-width");
   padding: design-token("space.1") design-token("space.2");
   color: component-token("banner", "color");
-  background-color: component-token("banner", "background.color");
+  background-color: component-token("banner", "color.background");
 
   @include breakpoint-sm-up {
     flex-direction: row;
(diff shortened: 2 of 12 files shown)


easy-ui-react/src/Banner/Banner.module.scss
```diff
@@ -0,0 +1,67 @@
...
+
+.variantNeutral {
+  @include component-token(
+    "banner",
+    "background_color",
+    theme-token("color.background.neutral")
+  );
+}
+
+.Banner {
```
ldewald, line 41: **Question:** Are we missing setting a max width and margin? There's a note in the Figma file that says

> Notes:
Banner is centered with 48px left and right padding. Banner width adjusts to take up rest of width with a max width of 1348
Text wraps and expands height if needed

easy-ui-react/src/Banner/Banner.module.scss
```diff
@@ -0,0 +1,67 @@
...
+  @include component-token(
+    "banner",
+    "border_radius",
+    design-token("shape.border_radius.base")
+  );
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  border-radius: 0;
```
ldewald, line 51: **Question:** Is this necessary? I think since we specify it for >tablet on line 61 this can be deleted.
```suggestion
```

easy-ui-react/src/Banner/Banner.module.scss
```diff
@@ -0,0 +1,69 @@
...
+  @include component-token("banner", "max-width", 1348px);
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  border-radius: 0;
+  flex-wrap: wrap;
+  white-space: pre-wrap;
+  max-width: component-token("banner", "max-width");
+  padding: design-token("space.1") design-token("space.2");
```
OskiTheCoder, line 56: will check with design about a few of these styling discrepancies: 

- padding numbers, the figma suggests `32px` horizontally on desktop, `16px` on tablet, and `24px` on mobile, but i imagine the latter two should be swapped.
- the colors for the `success` variant, i imagine its supposed to have `blue-800` text as is the case with the buttons, but it looks like `#000` in the spec. 
- overall height of the banner in the design spec suggests it should be `32px`, but i think thats intended to be `40px` unless im mistaken

easy-ui-react/src/styles/_media-queries.scss
```diff
@@ -0,0 +1,25 @@
...
+  @include breakpoint-sm-up {
+    @content;
+  }
+}
+
+@mixin ezui-desktop {
+  @include breakpoint-md-up {
+    @content;
+  }
+}
```
stephenjwatkins, line 25: **thought:** while i like these shortcuts, i feel a little on the fence with them because `tablet` and `desktop` seem like arbitrary tags to me. it's not clear to me what defines `tablet` and `desktop` in the real world. unless we get strict labeling from Design, while it is slightly less friendly, i would personally prefer using the `breakpoint-sm-up` and `breakpoint-md-up` since i think it's more explicit in aligning with our breakpoints

but curious how you feel after reading that

easy-ui-react/src/Banner/Banner.module.scss
OskiTheCoder: from what i can tell on the design spec, it seems like on mobile the border radius is 0 and takes up the full width, but will check in with design about this to be cerain

easy-ui-react/src/Banner/Banner.module.scss
OskiTheCoder: - with regards to outer margins, my thinking was that this kind of thing would be handled by a layout concerned component. steve and i briefly discussed it on the spec ticket here https://github.com/EasyPost/easy-ui/pull/87#discussion_r1165672389 , but on second thought, it does give me pause since this component is quite specific in its use. curious how you feel about this though, im open to ideas!

- with regards to `max-width`, i updated the PR to handle it, but there might be a more elegant way to deal with it, i tried setting it on the banner div itself, but that didn't seem to work. i ended up adding an outer div with these styles. open to ideas if you feel theres a better approach though

```css
.bannerContentContainer {
  display: flex;
  max-width: 1348px;
}

```

documentation/specs/Banner.md
```diff
@@ -22,13 +22,9 @@ export type BannerProps = {
    * @default "success"
    */
   status?: BannerStatus;
-  /**
-   * Banner emphasis text
-   */
+  /** Banner emphasis text */
   emphasis?: ReactNode;
```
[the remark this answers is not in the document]
OskiTheCoder: agree

easy-ui-react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,65 @@
...
+export type BannerProps = {
+  /**
+   * Banner variant color
+   * @default "success"
+   */
+  variant?: BannerVariant;
+  /** Banner emphasis text */
+  emphasis?: ReactNode;
+  /** Banner content */
+  children?: ReactNode;
```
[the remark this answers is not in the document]
OskiTheCoder: yep, it is indeed a valid use case to only render `emphasisText`

easy-ui-react/src/styles/_media-queries.scss
OskiTheCoder: i implemented your suggestion, but i had defined the mixin based on a spec in figma regarding breakpoints:

image

easy-ui-react/src/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  },
+  component: Banner,
+};
+
+export default meta;
+
+export const Controls: Story = {
+  render: Template.bind({}),
+  args: {
+    emphasis: "Banner text: ",
```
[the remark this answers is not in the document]
OskiTheCoder: i think i agree and i went ahead and implemented the suggestion, but let me know if i misunderstood the idea!

easy-ui-react/src/styles/_media-queries.scss
stephenjwatkins: oh i haven't seen this in Figma. when i last looked it was using the `xs`, `sm`, etc. interesting. "Web" doesn't quite make sense but i suppose desktop like you had it makes a reasonable alternative. it still feels slightly iffy to me but i'd be down to do it like this if you want. how do you think we would represent the different desktop sizes? t-shirt sizes? `desktop`, `desktop-lg`, `desktop-xl`, `desktop-2xl`, etc?

OskiTheCoder: hmm, i dont feel super strongly about including it and if youre iffy on it, id feel more comfortable if we leave it out of the mixin for now at least. probably something we can revisit when we're developing molecules or layout concerned components

easy-ui-react/src/Banner/Banner.module.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.variantSuccess {
+  @include component-token(
+    "banner",
+    "color",
+    theme-token("color.text.success.inverse")
+  );
+  @include component-token(
+    "banner",
+    "background.color",
```
stephenjwatkins, line 11: **nit:** in my previous review comment i was thinking this could be `color.background` to stick with our paradigm of `<foundation>.<property>.<modifiers>`

easy-ui-react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,76 @@
...
+ * When rendering a banner with both emphasisText and text via children, a delimeter ":" is
+ * automatically applied.
+ *
+ * This component is useful when wanting to inform users of new features, promotions and
+ * other offerings.
+ *
+ * @example
+ * Default success variant with emphasis and body:
+ * ```tsx
+ *  <Banner emphasis={<>Limited Time Only:</>}>
```
stephenjwatkins, line 40: **issue:** i think this needs to be updated to `emphasisText`

easy-ui-react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { ReactNode } from "react";
+import { Text } from "../Text";
+import { classNames, variationName } from "../utilities/css";
+import styles from "./Banner.module.scss";
+
+export type BannerVariant = "primary" | "success" | "neutral";
```
stephenjwatkins, line 6: **question:** does `success` make sense in this context?

i know i've been back and forth on the naming of these values in terms of consistency with the rest of the components like the `Button`, but thinking as a consumer rendering a banner as a promotional item, i'm not sure `success` would make sense since it's not the result of an action like other alerts.

it seems like the banners instead are different degrees of prominence. like the green is the most prominent, blue is the next most prominent, followed by the darker blue. which makes me wonder if it should take more of a hierarchical naming path like `<Banner variant="primary" />` (green); `<Banner variant="secondary" />` (blue); `<Banner variant="tertiary" />` (darkblue). or with a little more semantic: `<Banner variant="pivotal" />` (green); `<Banner variant="relevant" />` (blue); `<Banner variant="peripheral" />` (darkblue). those are probably too jargon-y...just brainstorming concepts.

what are your thoughts?

easy-ui-react/src/Banner/Banner.module.scss
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @include breakpoint-md-up {
+    padding: design-token("space.1") design-token("space.4");
+  }
+}
+
+.bannerContentContainer {
+  display: flex;
+  max-width: 1348px;
```
stephenjwatkins, line 72: **question:** was there a reason for creating a separate container for this instead of setting it on the `Banner` itself? i believe it should work fine on the root element instead. (it may be acting up with the left/right padding but once the `box-sizing` is set correctly to `border-box`, which i updated in a recent diff, it should be good)

stephenjwatkins, line 72: **nit:** is this value worth setting in a `component-token` at the top so that we can ensure it's a little more visible?

stephenjwatkins (commented):
looking good. only a few nits and one question about the prop name

easy-ui-react/src/Banner/Banner.tsx
OskiTheCoder: good point.

i do see the utility in using `primary` to map to the most used variation, but my concern is that i often feel `primary` refers to a color and it could potentially become confusing. im a bigger fan of the `pivotal` and `relevant` naming, but i went ahead and renamed `variant` to `color` so that maybe its a bit more clear what `primary`, `success` and `neutral` are referring to.  although if you do feel strongly about going to the `variant` naming you suggested with `pivotal` and `relevant`, im on board with that too.

easy-ui-react/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,76 @@
...
+          styles.Banner,
+          styles[variationName("variant", variant)],
+        )}
+      >
+        {emphasisText && (
+          <Text variant="subtitle1" alignment="center">
+            <strong className={classNames(styles.emphasisTextContent)}>
+              {!children ? <>{emphasisText}</> : <>{emphasisText}: </>}
+            </strong>
+          </Text>
```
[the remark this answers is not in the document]
OskiTheCoder: ah nice, this approach is much better, i went ahead and included `strong` in the `TextAs` type

easy-ui-react/src/Banner/Banner.tsx
stephenjwatkins: i still don't feel super great but i don't have answers lol. might as well just move forward

easy-ui-react/src/Text/Text.tsx
```diff
@@ -4,7 +4,16 @@ import { classNames, getComponentThemeToken } from "../utilities/css";
 
 import styles from "./Text.module.scss";
 
-export type TextAs = "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "span";
+export type TextAs =
```
stephenjwatkins, line 7: thanks

stephenjwatkins (approved):

PR #172, opened 2023-05-04
Create `<Menu />` component

stephenjwatkins:
- Added `<Menu />` component
- Added italic Poppins font in support of disabled item styles
- Various other fixes

Checklist:

- [x] Visuals complete
- [x] Storybook stories
- [x] Code complete
- [x] TSDoc
- [x] Theme/Design tokens audited
- [x] Spec up-to-date
- [x] Cross-browser checked (Chrome, Safari, Firefox)
- [x] Changeset
- [x] Unit tests


OskiTheCoder:
**question:** for the `width="fit-trigger"` variation, do you happen to know whats the use case for that one? im wondering if theres any potential issue with truncating text that would need to be readable; although im guess in such a case we wouldn't use the variation, but curious about the use case if you know of them

stephenjwatkins:
> **question:** for the `width="fit-trigger"` variation, do you happen to know whats the use case for that one? im wondering if theres any potential issue with truncating text that would need to be readable; although im guessing in such a case we wouldn't use the variation, but curious about the use case if you know of them

i agree that's a concern. i'll ask design. i was going off of the Figma spec that said to match the dropdown width

easy-ui-react/src/Menu/Menu.module.scss
```diff
@@ -0,0 +1,99 @@
...
+  );
+  @include component-token("menu", "shadow", theme-token("shadow.overlay"));
+  @include component-token("menu", "padding.x", design-token("space.2"));
+  @include component-token("menu", "color.border", theme-token("color.border"));
+
+  @include responsive-prop("menu", "width", width);
+
+  background: component-token("menu", "color.background");
+  border: 1px solid component-token("menu", "color.border");
+  border-radius: component-token("menu", "border_radius");
```
OskiTheCoder, line 38: **question**: out of curiosity is there a benefit to defining the border as `1px solid component-token("menu", "color.border");` and on the next line explicitly defining the `border-radius`?

easy-ui-react/src/Menu/useScrollbar.ts
```diff
@@ -0,0 +1,26 @@
...
+      },
+    },
+    defer: false,
+  });
+  useEffect(() => {
+    if (scrollRef.current) {
+      initialize({ target: scrollRef.current });
+    }
+  }, [initialize, scrollRef]);
+}
```
OskiTheCoder, line 26: 🎉

OskiTheCoder:
my guess is that this is just storybook being weird, but thought id mention it just in case: i noticed when i play around with the controls, i cant get the menu to pop up again unless i remount or reset the controls.


https://github.com/EasyPost/easy-ui/assets/44451175/69f15fbb-38f0-4d04-a975-6242e1236d1d

easy-ui-react/src/Menu/Menu.module.scss
stephenjwatkins: hm i'm not sure i understand the question. what's the alternative that you're thinking of?

stephenjwatkins:
> my guess is that this is just storybook being weird, but thought id mention it just in case: i noticed when i play around with the controls, i cant get the menu to pop up again unless i remount or reset the controls.

i noticed that too. i believe it is Storybook being weird. it seems as though after setting `isOpen`, the query param gets stuck in the URL and prevents the state from being set properly. i'll take a look though

easy-ui-react/src/Menu/Menu.module.scss
OskiTheCoder: derp, i just realized i confused `border-radius` and `border-width` lol, so ignore that question, but could we use the `border_width` token here?

stephenjwatkins: oh good call. we can

stephenjwatkins:
> i agree that's a concern. i'll ask design. i was going off of the Figma spec that said to match the dropdown width

removed `width="fit-trigger"` per conversation with design

OskiTheCoder (approved):
this is very very good!

PR #179, opened 2023-05-05
use UnstyledButton in Button, DropdownButton, and IconButton

OskiTheCoder:
- uses `UnstyledButton` in `Button`, `DropdownButton`, and `IconButton` to DRY up code and defer core button logic to the `UnstyledButton` component


easy-ui-react/src/Button/Button.tsx
```diff
@@ -59,11 +57,6 @@ export const Button = forwardRef<null, ButtonProps>((props, inRef) => {
   } = props;
 
   const ref = useRef(null);
```
stephenjwatkins, line 59: **suggestion:** you can let me know if i'm missing something, but i believe you can remove this `ref` since it's not being used for anything local to the component

stephenjwatkins (approved):
niiice

PR #192, opened 2023-05-10
Create pull request template

stephenjwatkins:
I looked into creating multiple PR templates and you have to use query parameters to specify the template when creating a pull request, which feels a little too complicated for now.

I think this PR template won't be relevant for all PRs, mostly for component ones, but we can always simply edit it when creating a PR if it's not relevant. So using it as the default makes sense to me for now. We can adjust as we use it.

Thoughts on the emojis? lol. I'm not super great at picking emojis to use so any input is welcome there.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,10 +1,10 @@
-## Changes
+## 📝 Changes
 
 Please provide a brief summary of the changes made and why they were made.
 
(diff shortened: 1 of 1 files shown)


ldewald (approved):

ldewald:
The emojis you picked lgtm!

PR #194, opened 2023-05-10
textfield spec

OskiTheCoder:
- `TextField` component spec

### DIFF
--- a/documentation/specs/TextField.md
+++ b/documentation/specs/TextField.md
@@ -1,0 +1,307 @@
+# `TextField` Component Specification
+
+## Overview
+
+The `TextField` component allows users to input text on a single line and provides the essentials to be used as a form control element. An Easy UI password input variation can be had by setting `variant="protected"`.
+
+### Use Cases
+
+- Use this component when building forms that require `text`, `email`, `password`, and `tel` inputs.
+- Can be used as the basis for a `search` input.
+
+### Features
+
+- The net effect of setting `variant` to `protected` is to get a PasswordInput component with an appropriate visibility icon.
+- Setting the `size` property also sets size for leftAlignedIcon and rightAlignedIcon: The size values map to Easy UI's token sizes for icons.
+- When `errorText` is supplied with `validationState="invalid"`, `helperText` is overriden.
+- Underlying input type can be set to `text`, `email`, `password`, `tel`, or `search`.
+- `isLabelVisuallyHidden` property can be used to visually hide the label.
+
+### Risks and Challenges
+
+- Accomodating various input types and variations while keeping a relatively simple and approachable API.
+- Ensuring the component remains accessible when label is visually hidden.
+
+### Prior Art
+
+- [Paste `<Input />`](https://paste.twilio.design/components/input)
+- [Polaris `<TextField />`](https://polaris.shopify.com/components/selection-and-input/text-field)
+- [Spectrum `<TextField />`](https://react-spectrum.adobe.com/react-spectrum/TextField.html)
+
+---
+
+## Design
+
+### API
+
+```ts
+import type { AriaTextFieldProps } from "react-aria";
+
+/**
+ * Note that this does not represent an exhaustive list of the props
+ * provided by AriaTextFieldProps, but they are notable properties that
+ * consumers may use on the TextField component
+ */
+export type ValidationState = "valid" | "invalid";
+
+type AriaTextFieldProps = {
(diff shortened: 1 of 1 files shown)


documentation/specs/TextField.md
```diff
@@ -0,0 +1,307 @@
...
+The `TextField` component allows users to input text on a single line and provides the essentials to be used as a form control element. An Easy UI password input variation can be had by setting `variant="protected"`.
+
+### Use Cases
+
+- Use this component when building forms that require `text`, `email`, `password`, and `tel` inputs.
+- Can be used as the basis for a `search` input.
+
+### Features
+
+- The net effect of setting `variant` to `protected` is to get a PasswordInput component with an appropriate visibility icon.
```
OskiTheCoder, line 14: i suppose we could use the input `type` prop to accomplish the same thing instead of having a `protected` variation, but i feel that since our password inputs have the clickable visibility icon, it makes a bit more sense to me to capture that info into a separate prop, but open to ideas

documentation/specs/TextField.md
```diff
@@ -0,0 +1,307 @@
...
+        </div>
+      )}
+      {showErrorText && (
+        <div {...errorTextProps}>
+          <Text>{errorText}</Text>
+        </div>
+      )}
+    </div>
+  );
+}
```
OskiTheCoder, line 176: anatomy here is more or less a very rough outline. most of the logic that we add will be related to styles, which _could_ end up impacting the underlying markup structure. either way i think its worth updating this during the implementation phase

documentation/specs/TextField.md
stephenjwatkins: i don't think i have a preference. your reasoning seems sound to me.

out of curiosity, would a `variant` ever be `protected` when the `type` is not `password`?

documentation/specs/TextField.md
```diff
@@ -0,0 +1,307 @@
...
+   */
+  isLabelVisuallyHidden?: boolean;
+  /** Error text that appears below input */
+  errorText?: ReactNode;
+  /** Helper text that appears below input */
+  helperText?: ReactNode;
+  /** Left aligned icon */
+  leftAlignedIcon?: IconSymbol;
+  /** Right aligned icon */
+  rightAlignedIcon?: IconSymbol;
```
stephenjwatkins, line 103: **question:** do we want to keep this naming consistent with our buttons (`iconAtStart`, `iconAtEnd`)?

documentation/specs/TextField.md
```diff
@@ -0,0 +1,307 @@
...
+
+export function Component() {
+  const [value, setValue] = useState("");
+  return (
+    <>
+      <TextField
+        label="Error"
+        validationState="invalid"
+        helperText="Some text" // will be overriden in the presence of an invalid state with error text
+        errorText="Some error text"
```
stephenjwatkins, line 284: **thought:** having both a `helperText` and `errorText` for the same positional slot tied to different states of the component feels just slightly odd to me, however it probably makes the most sense to do it this way, otherwise the consumer has to think about updating the slot based on validation logic

stephenjwatkins (approved):
nice. it's as if you've done this before

documentation/specs/TextField.md
```diff
@@ -0,0 +1,307 @@
...
+   * leftAlignedIcon and rightAlignedIcon. For instance, with `sm` size, the associated leftAlignedIcon
+   * or rightAlignedIcon size maps to the Easy UI icon size of 'sm'.
+   * @default 'md'
+   */
+  size?: TextFieldSize;
+  /**
+   * Visually hides the label, but keeps it accessible.
+   * @default false
+   */
+  isLabelVisuallyHidden?: boolean;
```
OskiTheCoder, line 95: per design, its a valid use case to have inputs without labels. from my understanding to keep it accessible, we can either visually hide the label (seems like we have a mixin in easy-ui already to do this) or add an `aria-label` attribute to the input (consumer provides the value).

documentation/specs/TextField.md
OskiTheCoder: ah good catch, yes i think we want that here

documentation/specs/TextField.md
OskiTheCoder: good question. i checked with design and there are no plans as of yet to want to support a password input that deviates from whats currently in figma, so i removed the `variant` property

PR #195, opened 2023-05-10
Create `<Badge />` spec

stephenjwatkins:


documentation/specs/Badge.md
```diff
@@ -0,0 +1,186 @@
...
+### API
+
+```ts
+type BadgeVariant =
+  | "primary"
+  | "secondary"
+  | "black"
+  | "inverse"
+  | "gray"
+  | "positive"
```
ldewald, line 47: **Question:** Are we using positive/negative elsewhere? I thought I had mostly seen use use success/danger

stephenjwatkins: you're right! updated it to reflect what we've been using already in the lib

documentation/specs/Badge.md
```diff
@@ -0,0 +1,186 @@
...
+   * Accessible label for badge if it differs from its content.
+   * Required for icon-only badges.
+   */
+  accessibilityLabel?: string;
+
+  /** Primary badge label */
+  children?: ReactNode;
+
+  /** Badge icon */
+  icon?: IconSymbol;
```
OskiTheCoder, line 62: **question**: do you know if the intention is to use the `sm` or `md` icon size?

ldewald (approved):
LGTM - simple enough to not require a review meeting

documentation/specs/Badge.md
stephenjwatkins: good catch. it will need a new icon size to accommodate the `20px` in the design. my thought was we can update `sm` to use `20px` and add a new size `xs` for `16px`. i don't believe we're using `sm` currently so i think that should be fine.

OskiTheCoder (approved):
nice!

PR #196, opened 2023-05-11
Fix VSCode import paths

stephenjwatkins:
## 📝 Changes

Added a default VSCode settings file that has some common settings useful to our project. It isn't the actual `settings.json` file but a template to be copied by the developer. This way it's not overwriting any custom user settings.

The actual fix for this problem is the setting:

```json
"typescript.preferences.importModuleSpecifier": "relative"
```

This setting tells VSCode to prefer relative imports, which is what we want for importing files from our own library. Note that VSCode still does absolute imports for libraries in `node_modules`.

Technically our TSConfig allows importing our components with an absolute path like it's been doing, and it will build with TSC like that, but our Storybook isn't configured to use the TSConfig in that way, and is why it breaks when developing against Storybook. We should maybe address that separately, but regardless of that working, it's better in our case to use relative imports anyway.


OskiTheCoder (approved):
🎉

PR #208, opened 2023-05-16
Create `<Badge />` component

stephenjwatkins:
## 📝 Changes

- Adds `<Badge />` component
- Adds `Anchor` icon to support Figma designs
- Updates design tokens for new icon and border radius sizes

## ✅ Checklist

- [x] Design reviewed
- [x] Ensure accessibility audit check
- [x] Changeset is added


documentation/specs/Badge.md
```diff
@@ -65,7 +65,7 @@ type Badge = {
    * Any additional text to support the primary label.
    * Only applicable on text-only badges.
    */
-  supportText?: ReactNode;
+  secondaryLabel?: ReactNode;
```
stephenjwatkins, line 68: seeing `supportText` in use wasn't super clear to me so i updated it to `secondaryLabel`. let me know if this is better or if i should revert or if there are better ideas.

easy-ui-icons/src/Anchor.json
stephenjwatkins, line 1: added for support to match Figma visuals. not sure it'll be in use for production but it shouldn't cause any harm if not

easy-ui-react/src/Badge/Badge.module.scss
```diff
@@ -0,0 +1,95 @@
...
+  color: component-token("badge", "secondary.color.text");
+}
+
+// Wrapper element around the text to normalize the hor and ver padding
+.text {
+  display: inline-flex;
+  padding: design-token("space.0.25") design-token("space.0.5");
+}
+
+// prettier-ignore
```
stephenjwatkins, line 33: reads a little better in IDEs as single lines. although admittedly when wrapping is enabled (like here) it reads less friendly

easy-ui-react/src/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+  },
+};
+
+export const WithTooltip: Story = {
+  render: (args) => (
+    <Tooltip content="Here is a tooltip">
+      <span tabIndex={0} style={{ display: "inline-flex", borderRadius: 8 }}>
+        <Badge {...args} />
+      </span>
+    </Tooltip>
```
stephenjwatkins, line 121: this use case feels a little awkward admittedly. not sure how common it will be. tooltips require focusable elements and the Badge is not natively focusable, so an intermediate element is created to provide focus. i wanted to not add too much tooltip-specific code inside Badge since they should be fairly independent. let me know any thoughts or improvements.

easy-ui-react/src/Tooltip/Tooltip.tsx
```diff
@@ -124,7 +124,10 @@ export function Tooltip(props: TooltipProps) {
 
   return (
     <>
-      {React.cloneElement(children, { ...triggerProps, ref: triggerRef })}
+      {React.cloneElement(children, {
+        ...mergeProps(triggerProps, children.props),
```
stephenjwatkins, line 128: `useTooltipTrigger()` from React Aria sets some `triggerProps` to `undefined`, like `tabIndex`, which overrides values provided by the consumer. this ensures props provided by the consumer are merged into the trigger props provided

easy-ui-tokens/src/shape/base.json
```diff
@@ -4,6 +4,10 @@
       "base": {
         "value": "5px",
         "attributes": { "unit": "rem" }
+      },
+      "lg": {
```
stephenjwatkins, line 8: do t-shirt sizes make sense?

easy-ui-tokens/src/theme/base.json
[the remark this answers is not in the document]
OskiTheCoder: yeah i had quite a hard time coming up with names for these too lol. i think your choices are sensible and ill adopt them into my ongoing PR for notifications.

**question**: is it okay that modifiers can map to different color shades? for ex, as far as i know `yellow-100` and `yellow-200` aren't in scope/use in any of our designs, so the `subtle` modifier i think it makes sense to use for  `yellow-300` and the other `<color>-100` shades, but i notice we also use it to map to `color.gray.050` for `black`. im probably overthinking it and it may not even be good practice to try to force a pattern when we really should strive for names that fit the situation and provide accurate context, which i think is what you did here, but curious if you have any additional thoughts.

easy-ui-react/src/Badge/Badge.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  it("should warn on missing children or icon", () => {
+    render(<Badge secondaryLabel="Secondary text" />);
+    expect(consoleWarnSpy).toBeCalledWith(
+      expect.stringMatching(/requires one of children or icon/),
+    );
+  });
+
+  it("should warn with no accessibility label on icon", () => {
```
OskiTheCoder, line 57: nice i like this test

documentation/specs/Badge.md
OskiTheCoder: i think `secondaryLabel` is good

easy-ui-tokens/src/theme/base.json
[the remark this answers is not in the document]
stephenjwatkins: > is it okay that modifiers can map to different color shades?

i had the same question when i was adding these values. i'm admittedly not entirely convinced one way or another. it can be confusing scanning the list of tokens and seeing inconsistency in the numbers. however, a benefit of having alias/theme tokens are that they can assign context to primitive values so that a "subtle" color in one shade can use a different lightness value than a "subtle" color in another shade. so i do think it's sticking to the spirit of that idea even if initially a little jarring

easy-ui-tokens/src/shape/base.json
stephenjwatkins: i thought the same thing. i went ahead and updated it to `md`.

easy-ui-react/src/Tooltip/Tooltip.tsx
OskiTheCoder: **question**: do you know if its coming from their `useFocusable` hook or somewhere else?

OskiTheCoder (approved):
nice! looks great

easy-ui-react/src/Tooltip/Tooltip.tsx
stephenjwatkins: yes i believe it's coming from `useFocusable`. i looked at the code and the intention is to simply "reset" `tabIndex` by assigning it to `undefined` but because that comes back as a key on the returned props it doesn't allow for proper merging with `cloneElement()`. another little quirk with react aria

PR #214, opened 2023-05-17
create notification component

OskiTheCoder:
## 📝 Changes

Please provide a brief summary of the changes made and why they were made.

Include any notes, screenshots, or videos that may be helpful for developers reviewing this pull request.

## ✅ Checklist

- [ ] Visuals are complete and match Figma
- [ ] Code is complete and in accordance with our style guide
- [ ] Design and theme tokens are audited for any relevant changes
- [ ] Unit tests are written and passing
- [ ] TSDoc is written or updated for any component API surface area
- [ ] Stories in Storybook accompany any relevant component changes
- [ ] Ensure no accessibility violations are reported in Storybook
- [ ] Specs and documentation are up-to-date
- [ ] Cross-browser check is performed (Chrome, Safari, Firefox)
- [ ] Changeset is added


easy-ui-react/package.json
```diff
@@ -31,7 +31,9 @@
   "dependencies": {
     "@easypost/easy-ui-icons": "1.0.0-alpha.2",
     "@easypost/easy-ui-tokens": "1.0.0-alpha.1",
+    "@react-aria/toast": "^3.0.0-beta.0",
```
OskiTheCoder, line 34: unless im missing something, it seems like the `toast` hooks aren't part of their monopackage yet for react-aria: https://github.com/adobe/react-spectrum/blob/main/packages/react-aria/src/index.ts my guess is that the `toast` hooks are still in beta

easy-ui-react/src/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+type Story = StoryObj;
+
+const meta: Meta<typeof Notification> = {
+  title: "Components/Notification",
+  argTypes: {},
+};
+
+export default meta;
+
+export const Notifications: Story = {
```
OskiTheCoder, line 21: as mentioned, each story gets its own `Provider` and hence their own queue. i found it best to group them in this way as to not pollute the console with warnings and to simulate the desired behavior that a new notification should end the cycle of the previous one. also, with regards to stacking and "priority",  based on my latest discussion, one notification at a time is what we want, but i will confirm/check again with design.

easy-ui-react/src/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+          left: "0px",
+          right: "0px",
+          bottom: "0px",
+        }}
+      >
+        <SimulatedNotificationTrigger message="Hello World" />
+      </EasyUIProvider>
+    );
+  },
+};
```
OskiTheCoder, line 125: so technically, if someone issues a notification with the `Notification` story above and then comes to this story to issue another notification while the other is still visible, a warning is logged to the console: 

`Page contains more than one landmark with the 'region' role and 'Notifications' label. If two or more landmarks on a page share the same role, they must have unique labels`

but im more willing to accept this scenario since this story highlights the custom placement aspect

easy-ui-react/src/Notification/Notification.mdx
```diff
@@ -0,0 +1,80 @@
...
+
+<Canvas of={NotificationStories.Notifications} />
+
+## Custom placement
+
+Notifications can have custom placement on a per app basis. Use the `notificationPlacementOffset` prop which is available on the `EasyUIProvider` component.
+
+<Canvas of={NotificationStories.CustomPlacement} />
+
+## Properties
```
OskiTheCoder, line 23: i know we've been using `ArgsTable` here, but with the base `Notification` component, the output included the `state` and `toast` prop, which i think has the potential to become a little confusing since those aren't relevant to consumers, but open to ideas/thoughts

easy-ui-tokens/src/theme/base.json
```diff
@@ -33,10 +49,29 @@
     "color.text.disabled": { "value": "{color.gray.300}" },
     "color.text.heading": { "value": "{color.blue.800}" },
     "color.text.inverse": { "value": "{color.gray.000}" },
-    "color.text.success.inverse": { "value": "{color.blue.800}" },
-    "color.text.success.inverse.pressed": { "value": "{color.gray.000}" },
+    "color.text.black.inverse.bold": { "value": "{color.gray.000}" },
+    "color.text.black.inverse.subtle": { "value": "{color.blue.800}" },
+    "color.text.danger.bold": { "value": "{color.red.800}" },
```
OskiTheCoder, line 54: i added some new text colors, but tbh, idk how i feel about having a token text color name with `color.text.danger.bold`, open to ideas on it though. i thought about maybe creating another set of background tokens, something like `"color.background.<color>.notification"` and the corresponding text would just follow the inverse pattern, but it maybe felt a little too specific lol

easy-ui-react/src/Notification/Notification.module.scss
```diff
@@ -0,0 +1,83 @@
...
+
+.notificationRegion {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.notificationContainer {
+  position: fixed;
+  z-index: 999999;
```
stephenjwatkins, line 80: **suggestion:** i think z-indices would be a good thing to tokenize, that way we have a reference for all values used. would probably good to have a base layer of tokens for primitives like `z-index-999999`; and then probably an alias token layer that references the primitive for specific contexts like `z-index-notification: z-index-999999`

easy-ui-react/src/Notification/Notification.module.scss
```diff
@@ -0,0 +1,83 @@
...
+.closeButton {
+  border: 0;
+  padding: design-token("space.0");
+  margin: design-token("space.0.5") design-token("space.1") design-token("space.0.5") design-token("space.2");
+  background-color: component-token("notification", "color.background");
+  color: component-token("notification", "color.text");
+  cursor: pointer;
+}
+
+.notificationRegion {
```
stephenjwatkins, line 72: **nit:** since notification is the name of the component would it make sense for these to be more brief and just be `region` and `container`?

easy-ui-react/src/Notification/Notification.module.scss
```diff
@@ -0,0 +1,83 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.notificationContainer {
+  position: fixed;
+  z-index: 999999;
+  width: 100%;
+  top: 0;
```
stephenjwatkins, line 82: **issue:** without a `left: 0;` property, the notification will be left-aligned to whatever the body padding is, which may not be what we want.

i notice in Storybook that because there is no `left: 0` it exceeds past the story window and cuts off the right-most edge:

image

easy-ui-react/src/Notification/Notification.module.scss
```diff
@@ -0,0 +1,83 @@
...
+  background-color: component-token("notification", "color.background");
+}
+
+.iconText {
+  display: inline-flex;
+  align-items: center;
+  gap: design-token("space.1");
+}
+
+// prettier-ignore
```
stephenjwatkins, line 62: **nit:** would it make sense for some of these cases to be a little more selective with the `// prettier-ignore`?

easy-ui-react/src/Notification/Notification.module.scss
```diff
@@ -0,0 +1,83 @@
...
+  @include component-token("notification", "max-width", 1200px);
+  align-items: center;
+  justify-content: space-between;
+  margin: auto;
+  padding: design-token("space.1");
+  border-radius: 0 0 component-token("notification", "border_radius") component-token("notification", "border_radius");
+  max-width: component-token("notification", "max-width");
+  color: component-token("notification", "color");
+  background-color: component-token("notification", "color.background");
+}
```
stephenjwatkins, line 54: **issue:** i notice that when a toast is open on the story that the `container` and `region` are blocking the cursor from hitting elements underneath it. probably we want `pointer-events: none;` on the container and `pointer-events: auto;` on the `.Notification`

when the toast is open below the tops of some of the buttons can't be clicked due to the container elements cutting off the cursor:

image

easy-ui-react/src/Notification/Notification.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+  title: "Components/Notification",
+  argTypes: {},
+};
+
+export default meta;
+
+export const Notifications: Story = {
+  render: () => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const { notification } = useNotification();
```
stephenjwatkins, line 24: **question:** is it "ok" to have hooks in this render function?

easy-ui-react/src/Notification/Notification.stories.tsx
stephenjwatkins: i wasn't able to replicate the console warning but does it do this because it's wrapped in multiple providers, one at the parent and one here?

easy-ui-tokens/src/theme/base.json
stephenjwatkins: i think for now this is good. it seems to mirror what other design systems are doing with their alias tokens.

i think we can have a retro on this after we're done with our atoms to see how the patterns aged

easy-ui-react/src/Notification/NotificationRegion.tsx
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function NotificationRegion(props: NotificationRegionProps) {
+  const { notification } = props;
+  const ref = React.useRef(null);
+  const { regionProps } = useToastRegion(props, notification, ref);
+  /**
+   * `visibleToasts` and `toast` are artifacts of react-stately and in
+   * Easy UI `visibleToasts` is being treated as `visibleNotifications` and
+   * similarly, `toast` here is being treated as a `notification`.
+   *
```
stephenjwatkins, line 31: **nit:** extra line

easy-ui-react/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,344 @@
...
+*```
+ */
+export function useNotification() {
+  const notification = useContext(NotificationContext);
+  if (!notification) {
+    throw new Error(
+      "useNotification() must be used within a NotificationProvider, which is bundled with EasyUIProvider",
+    );
+  }
+  return { notification: notification };
```
stephenjwatkins, line 305: **question:** what other properties are returned on `notification` other than the `show*Toast` and `show*Alert` methods given that it also extends `ToastState<NotificationProps>` as well? follow up is do we want to expose that to the consumer?

stephenjwatkins, line 305: **question:** would it make sense to expose this as just `return notification;` instead of as an intermediary object `return { notification };`?

```tsx
const notification = useNotification();
const onToast = () => { notification.showToast(); }
```

or are you designing it to be extensible in the future?

stephenjwatkins (commented):
lot of nice work in here. a few minor comments.

one thing i'm realizing with the `position: fixed` approach for the notification container is that it doesn't play well with aligning to our sticky header.

notice how with `fixed` positioning on an element under the body that it doesn't adhere to the stuck header:

https://github.com/EasyPost/easy-ui/assets/752942/c03260c9-01a4-44ec-a03a-d024193b4092

image

vs being attached with `position: absolute` to the header:

https://github.com/EasyPost/easy-ui/assets/752942/c8001f06-c5ee-49cb-911e-217fa8ee98a1

image

i didn't have this realization in the spec-ing process unfortunately. the behavior with the `absolute` positioning feels a bit more robust than the `fixed` approach, but of course this would require a little modification to the way the container works. not sure it's a blocker but wanted to share in case it sparks any immediate ideas

easy-ui-react/src/Notification/Notification.stories.tsx
OskiTheCoder: it comes from `react-aria` `useLandmark` [hook](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/landmark/src/useLandmark.ts#L217) which is used behind the scenes by `useToastRegion`.

easy-ui-react/src/Notification/Notification.module.scss
OskiTheCoder: i think this is a good call. i also added `absolute` positioning to the container

easy-ui-react/src/Notification/Notification.module.scss
OskiTheCoder: ah interesting, didn't know that, added it!

easy-ui-react/src/Notification/Notification.module.scss
OskiTheCoder: if i understand correctly, i added `/* prettier-ignore */` instead of `// prettier-ignore`, which i think has block scope, but let me know if i misunderstood the suggestion

easy-ui-react/src/Notification/Notification.tsx
OskiTheCoder: my initial thought was that we do want to include some of the properties included by `ToastState` [here ](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/toast/src/useToastState.ts#LL48C33-L48C33) but after thinking about it a bit more, i think it has the potential to become unnecessarily confusing for consumers having the `show<Status><Type>` methods alongside a generic `add`. i was thinking maybe theres a use case for needing programmatic dismissal and having the ability to pause toasts, but again, i think it could become too confusing. i updated the PR so that we're only exposing the `show<Status><Type>` methods.

OskiTheCoder: agree, it does make sense to expose it just as notification instead of an intermediary object.

easy-ui-react/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,374 @@
...
+
+  const notification = {
+    showErrorAlert: combinedState.showErrorAlert,
+    showErrorToast: combinedState.showErrorToast,
+    showPromotionalToast: combinedState.showPromotionalToast,
+    showPromotionalAlert: combinedState.showPromotionalAlert,
+    showNeutralToast: combinedState.showNeutralToast,
+    showNeutralAlert: combinedState.showNeutralAlert,
+    showSuccessToast: combinedState.showSuccessToast,
+    showSuccessAlert: combinedState.showErrorAlert,
```
stephenjwatkins, line 360: looks like this got mapped wrong:

```suggestion
    showSuccessAlert: combinedState.showSuccessAlert,
```

image

easy-ui-react/src/Notification/Notification.module.scss
stephenjwatkins: i think the way prettier ignore works is that it ignores the block that it's attached to. so if it's above a selector it'll apply to the whole block and if it's just above a line it'll apply to that line. so i was just meaning we could selectively apply it to lines instead of whole blacks to limit the scope of the ignore. but seems like it's pretty good now

easy-ui-react/src/Notification/Notification.tsx
stephenjwatkins: got ya. i agree regarding `add`. i think it would be confusing being in addition to our `show` methods. however i also agree that programmatic dismissal may come in handy.

a specific use case i'm wondering about: in apps, many times it's desired to close all toasts on page navigation. i'm wondering how that could be accomplished with what we have exposed?

easy-ui-react/src/Notification/Notification.module.scss
stephenjwatkins: the problem with using `absolute` on an element that's connected to `document.body` is that it will move with the scroll, so if a user is scrolled down on the page and a toast appears it may be above what's visible on their screen. so `fixed` position is the better positioning to use when attached to the body.

`absolute` positioning would work if it was a child of the sticky header in our easypost-web app but that would require the consumer to be in control of the container. i guess one option would be allowing for a custom portal container to be passed in along with "fixed" or "absolute" being passed in to the `notificationPlacementStyle`, etc

easy-ui-tokens/src/z-index/base.json
```diff
@@ -0,0 +1,7 @@
+{
+  "z-index": {
+    "999999": {
+      "value": "999999"
+    }
+  }
+}
```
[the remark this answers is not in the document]
OskiTheCoder: updated token, let me know if this is roughly what you had in mind or if we should create a separate file in `z-index` that reads from the value and has the alias naming

easy-ui-react/src/Notification/Notification.tsx
OskiTheCoder: oops, good catch

easy-ui-react/src/Notification/Notification.tsx
OskiTheCoder: thats a good point, i went and added the ability to programmatically dismiss notifications via `closeActiveNotification()`

on navigation on our setup, it should be happening as is without needing to do it programmatically

https://github.com/EasyPost/easy-ui/assets/44451175/219993fe-d0da-4c5e-ae05-d0bb38f936b1

easy-ui-react/src/Notification/Notification.module.scss
OskiTheCoder: if i understand the suggestion correctly, i refactored the logic a bit for the `Provider` to accept a generic `notificationPlacementProps` prop which is an object that can accept an `htmlId`, `positionType`, and `positionPlacement` prop. if `htmlId` is provided, then we render the notifications into that container instead of the `document.body`

```tsx
export type NotificationPositionPlacement = {
  /** Top offset */
  top?: string;
  /** Right offset */
  right?: string;
  /** Bottom offset */
  bottom?: string;
  /** Left offset */
  left?: string;
};

export type NotificationPositionType = "fixed" | "absolute";

export type NotificationPlacementProps = {
  /** HTML ID of element where notifications will render to */
  htmlId?: string;
  /** Position type */
  positionType?: NotificationPositionType;
  /** Position placement */
  positionPlacement?: NotificationPositionPlacement;
};

```

easy-ui-react/src/Notification/Notification.tsx
```diff
@@ -0,0 +1,391 @@
...
+  positionType?: NotificationPositionType;
+  /** Position placement */
+  positionPlacement?: NotificationPositionPlacement;
+};
+
+export type NotificationProviderProps = {
+  /** Child components */
+  children: ReactNode;
+  /** Notification placement props */
+  notificationPlacementProps?: NotificationPlacementProps;
```
stephenjwatkins, line 344: **suggestion:** nice i like this architecture

just a couple suggestions you can feel free to take or leave:

- allow for accepting an element instead of an id, this way we don't have to presuppose anything about the consumer. they don't have to give the element an id for instance. and it's easy enough to just do `document.getElementById()` on their end or even pass a React `ref.current`
- consider making some names a little more brief

```tsx
<EasyUIProvider
  notificationPlacement={{
    container: document.querySelector('.epr-header'),
    position: "absolute",
    offset: { top: 0, left: 0 }
  }}
>
```

OskiTheCoder: - i opted for a `containerFn` prop where the consumer could pass a simple function that gets the element. the idea is when the consumer wants to pass a container, ideally they want the specific styles they pass along to apply as well, but if the container doesn't exist or they pass a bad prop, we have control over the request to default to the original styles if the request fails (of course this could be done another way too). also, we get the added benefit of only issuing the call to get the container when necessary via the `state` prop. let me know if you think theres a downside to this approach though

```tsx
<EasyUIProvider
  notificationPlacement={{
    containerFn: () => document.querySelector('.epr-header'),
    position: "absolute",
    offset: { top: 0, left: 0 }
  }}
>
```

- updated names to be more brief

stephenjwatkins: works for me. i think i typically see those functions named as `getContainer: () => document...` but what you have is fine too.

did we want to update to `notificationPlacement` instead of `notificationPlacementProps`? i prefer slightly more brief but up to you

stephenjwatkins (approved):
niiice

PR #238, opened 2023-05-24
Create `<Card />` spec

stephenjwatkins:
## 📝 Changes

Create spec for `<Card />`.

Went through quite a few permutations locally and landed on what I feel is the best in terms of offering something simple for common use cases but allowing for more advanced composability for more complex use cases.


documentation/specs/Card.md
```diff
@@ -0,0 +1,194 @@
+# `Card` Component Specification
```
OskiTheCoder, line 1: **question**: does the checked card variation thats shown in figma have any significance to the content that the card renders or is it just a use case example and not really a stateful variation of some sorts?

documentation/specs/Card.md
```diff
@@ -0,0 +1,194 @@
...
+
+Card split into two columns using `HorizontalGrid` and deconstructed Card pieces:
+
+```tsx
+import { Card } from "@easypost/easy-ui/Card";
+
+function Component() {
+  return (
+    <Card.Container {...args}>
+      <HorizontalGrid columns={2}>
```
OskiTheCoder, line 132: **question**: do you envision the `HoriztonalGrid` to be a general purpose layout component?

stephenjwatkins: yeah it would be general purpose layout primitive, along with `VerticalStack`. both from Polaris. they both seem pretty solidly designed and useful.

i actually like a lot of layout components that polaris have, at least their new ones:

https://polaris.shopify.com/components/layout-and-structure

i imagine we may end up having most of them

stephenjwatkins:
talked to Rachel and updated the API a little bit per our discussion. feeling pretty good where things are at. the deconstructed pieces of `<Card />` are helpful for composability and we can create lightweight wrappers around them for common cases that we don't want consumers to have to repeat, like a `<CheckboxCard />`

documentation/specs/Card.md
```diff
@@ -0,0 +1,247 @@
...
+## Design
+
+A `Card` is semantically straightforward. Essentially serving as a container with preset styles. The complexity is surfacing a sufficiently advanced API that can accommodate the use cases without becoming overburdened.
+
+Architecture proposed is to surface a basic `<Card />` component with `variety` and `background` props while exposing the deconstructed parts to allow for maximum composability for complex use cases. The underlying pieces will be the `<Card.Container />` which houses the `variant` prop for determining the overall style of the card, and then a `<Card.Area />` which contains the inner container properties such as padding and the `background`.
+
+### API
+
+```ts
+export type CardAs = "div" | "label" | "button" | "a" | "fieldset";
```
OskiTheCoder, line 38: **question**: is the idea to use the `UnstyledButton` to handle the `"button"` and `"a"` scenario?

OskiTheCoder (approved):
lgtm! nice

PR #244, opened 2023-05-26
textfield component

OskiTheCoder:
## 📝 Changes

- Adds `TextField` component

## ✅ Checklist

- [x] Changeset is added


easy-ui-react/src/TextField/TextField.module.scss
```diff
@@ -0,0 +1,274 @@
+@use "../styles/common" as *;
```
OskiTheCoder, line 1: need to check in with design about a few things:

-need to check if the error state / flow aligns with expectations
- i dont think this padding configuration is possible unless the one on the right is referring to the input itself having difference text alignment:

input

- for the icon cases, need to confirm/check padding alignment.
- storybook is reporting an accessibility violation regarding color contrast with the optional text.
- not sure if small input variations can have label text with emphasis

easy-ui-react/src/TextField/TextField.stories.tsx
```diff
@@ -0,0 +1,210 @@
...
+export const VisuallyHiddenLabel: Story = {
+  render: Template.bind({}),
+  args: {
+    label: "Label",
+    isLabelVisuallyHidden: true,
+  },
+  decorators: [InputDecorator],
+};
+
+export const Controls: Story = {
```
OskiTheCoder, line 139: couldn't get storybook to document the props correctly, hence the manual work for the controls story. it seems as though its having issues with the union between `TextFieldProps` and `AriaTextFieldProps`. when i remove `AriaTextFieldProps`, the props get documented correctly, but with the union (as we need it), it was only picking up the props unique to `TextFieldProps`.

easy-ui-react/src/TextField/TextField.module.scss
```diff
@@ -0,0 +1,274 @@
...
+  );
+  outline: none;
+  position: relative;
+  width: 100%;
+
+  &:focus,
+  &:active {
+    color: component-token("textfield", "color.text.engaged");
+    border: design-token("shape.border_width.1") solid
+      component-token("textfield", "color.border.engaged");
```
stephenjwatkins, line 101: **suggestion:** if the border size and style doesn't change from what's set on the parent selector, it'd be preferable to only update the part that changes. in this case i believe it's just `border-color`. that way you only have one place where the primary border configuration is set

easy-ui-react/src/TextField/TextField.module.scss
```diff
@@ -0,0 +1,274 @@
...
+  border-left: none;
+  margin: 0;
+  padding: calc(
+    #{design-token("space.1.5")} - #{design-token("shape.border_width.1")}
+  );
+  padding-left: design-token("space.1.5");
+
+  &:focus,
+  &:active {
+    border: design-token("shape.border_width.1") solid
```
stephenjwatkins, line 173: **suggestion:** same comment here as the previous with repeating border configurations

easy-ui-react/src/TextField/TextField.module.scss
```diff
@@ -0,0 +1,274 @@
...
+  }
+}
+
+.passwordBtnError {
+  border-color: component-token("textfield", "color.border.danger");
+  &:focus,
+  &:active {
+    border-color: component-token("textfield", "color.border.danger");
+  }
+}
```
stephenjwatkins, line 198: **suggestion:** you could reduce the border-color declarations by combining selectors
```suggestion
.passwordBtnError,
.passwordBtnError:focus,
.passwordBtnError:active {
  border-color: component-token("textfield", "color.border.danger");
}
```

easy-ui-react/src/TextField/TextField.stories.tsx
stephenjwatkins: hm yeah that's unfortunate. this seems like a lot of effort to write/maintain

easy-ui-react/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,359 @@
...
+    label,
+    errorText,
+    helperText,
+    placeholder,
+    value,
+    defaultValue,
+    iconAtStart,
+    iconAtEnd,
+  } = props;
+  const [isVisible, setIsVisible] = useState(false);
```
stephenjwatkins, line 201: **nit:** the name `isVisible` wasn't immediately obvious to me what it represented since we have other concepts around visibility (like label being visually hidden). it's not a big deal if there isn't a better name, it may be fine as-is, but wondering if something else would be more clear like `isPasswordVisible` or `isInputValueVisible`

easy-ui-react/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  const isPassword = type === "password";
+  const hasError = validationState === "invalid";
+  const showErrorText = hasError && errorText;
+  const showHelperText = !showErrorText && helperText;
+  const canUseIcon = !bothIconPropsDefined && !isPassword;
+  const hasStartIcon = canUseIcon && iconAtStart;
+  const hasEndIcon = canUseIcon && iconAtEnd;
+  const typeAdjustedForPasswordVisibility = isPassword && isVisible;
```
stephenjwatkins, line 224: **nit:** since this is a boolean could we name it `isTypeAdjustedForPasswordVisibility`?

easy-ui-react/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,359 @@
...
+    <div className={classNames(styles.root)}>
+      <div className={styles.inputLabelContainer}>
+        <label
+          {...labelProps}
+          className={classNames(
+            styles.label,
+            isLabelVisuallyHidden && styles.labelHidden,
+          )}
+        >
+          {getLabelText(
```
stephenjwatkins, line 236: **question:** would it be useful for `getLabelText` and `getIcon` to be components instead of functions? `<LabelText />` and `<InputIcon />`

not a big deal but React brings along some niceties around invoking components rather than pure functions that can be nice to take advantage of whenever we can

easy-ui-react/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,359 @@
...
+          </Text>
+        </div>
+      )}
+      {showErrorText && (
+        <div {...errorTextProps} className={styles.caption}>
+          <Text variant="caption" color="danger">
+            {errorText}
+          </Text>
+        </div>
+      )}
```
stephenjwatkins, line 299: **question:** it may be overkill so it's up to your discretion, but would it be worth creating a component that these each could use to eliminate the duplicities? `<TextFieldCaption />`

easy-ui-react/src/TextField/TextField.tsx
```diff
@@ -0,0 +1,359 @@
...
+              )}
+              onPress={() => setIsVisible((prevVisibility) => !prevVisibility)}
+              isDisabled={isDisabled}
+            >
+              <Text visuallyHidden>password visibility</Text>
+              <Icon
+                symbol={isVisible ? VisibilityIcon : VisibilityOffIcon}
+                size={mapIconSize(size)}
+              />
+            </UnstyledButton>
```
stephenjwatkins, line 280: **question:** it may be overkill so up to you, but would it be worth creating a component for this to keep `TextField` main render body a little more streamlined? `<TextFieldPasswordButton />` or w/e

stephenjwatkins (commented):
overall looks good. a couple thoughts noted in comments.

one other general comment is that the `TextField` and `TextField.module.scss` files feel a little heavy on code, which can be a little daunting to maintain in the future. it's honestly not that bad, and certainly doesn't need to be a blocker, but if you happen to think of a way to break it out a little to more manageable pieces, i think our future selves would be a little more happy with making modifications to it

easy-ui-react/src/TextField/TextField.module.scss
OskiTheCoder: ah yeah, good call

easy-ui-react/src/TextField/TextField.module.scss
OskiTheCoder: i think in this case i do want the whole border defined, but let me know if im missing something. in the resting state, i have 

```css
  border: design-token("shape.border_width.1") solid
    component-token("textfield", "color.border.resting");
  border-left: none;
```

if i only define `border-left`, ill miss the `active/focus` `border-color` styles that need to apply all around. similarly, if i only define `border-color`, the `border-left` wont exist.

easy-ui-react/src/TextField/TextField.module.scss
OskiTheCoder: good call

easy-ui-react/src/TextField/TextField.tsx
OskiTheCoder: updated to `isPasswordVisible`

easy-ui-react/src/TextField/TextField.tsx
OskiTheCoder: my initial hesitation for breaking it out was that it might make it more difficult to follow some of the logic, but on second thought, i think some of the concerns are separate enough that it does make sense to break it up a bit. also i realized the label logic will probably be needed again for the upcoming `TextArea` component. i went ahead and broke it up into `Label`, `InputIcon`, `InputCaption`, and `PasswordButton`. i think its a bit easier to read now, but let me know how it looks/feels to you

easy-ui-react/src/TextField/TextField.tsx
OskiTheCoder: created an `InputCaption` component; i think it will see some reuse in the `TextArea` component as well

easy-ui-react/src/TextField/TextField.tsx
OskiTheCoder: I think its worth it, created a `PasswordButton` component

easy-ui-react/src/TextField/TextField.module.scss
stephenjwatkins: got ya. i missed the `border-left`

stephenjwatkins (approved):

PR #274, opened 2023-06-05
create text area component

OskiTheCoder:
## 📝 Changes

- creates `<TextArea />` component

## ✅ Checklist

- [x] Changeset is added


easy-ui-react/src/TextField/TextField.module.scss
```diff
@@ -1,6 +1,8 @@
 @use "../styles/common" as *;
 
 .root {
+  @include component-token("textfield", "min-height-textarea", 48px);
+  @include component-token("textfield", "min-height-textarea-lg", 56px);
```
OskiTheCoder, line 5: figma suggests min height of `32px` for the standard size, but i think its intended to be `48px`, will check/confirm though

easy-ui-react/src/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,126 @@
...
+      label={label}
+      errorText={errorText}
+      helperText={helperText}
+      placeholder={placeholder}
+      value={value}
+      defaultValue={defaultValue}
+      rows={rows}
+    />
+  );
+}
```
OskiTheCoder, line 126: the specs suggest there should be a custom scrollbar and as far as i know, for a textarea element this can be done via the `webkit-scrollbar` CSS pseudo-element (for FF i think we'd have to use `scrollbar` ?), but [MDN suggests](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar) we shouldn't do this and im not aware of a safe/reliable alternative, but maybe im missing something

easy-ui-react/src/TextField/useTextFieldElement.ts
```diff
@@ -0,0 +1,46 @@
+import { MutableRefObject } from "react";
+import { useTextField } from "react-aria";
+import { TextFieldProps } from "./TextField";
+
+export function useTextFieldElement(
```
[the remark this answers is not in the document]
stephenjwatkins: yeah, we probably don't want to call the hook twice. from what i understand, Aria has some global state that relies on a predictable calling order for the hooks, like wiring up IDs on the client and server

if we have to appease TypeScript with a hard cast that'd be preferable to breaking the runtime code

OskiTheCoder:
will reopen in a sec

PR #278, opened 2023-06-06
Create `<Toggle />` component

stephenjwatkins:
## 📝 Changes

- Creates toggle component

## ✅ Checklist

- [x] Changeset is added


easy-ui-tokens/src/theme/base.json
```diff
@@ -8,11 +8,14 @@
     "color.background.danger.hovered": { "value": "{color.red.600}" },
     "color.background.danger.pressed": { "value": "{color.red.700}" },
     "color.background.danger.subtle": { "value": "{color.red.100}" },
+    "color.background.disabled.subtler": { "value": "{color.gray.050}" },
```
stephenjwatkins, line 11: 😆 looking forward to auditing these

easy-ui-react/src/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,134 @@
...
+      {children && (
+        <span className={styles.text}>
+          <Text variant="body1" color={textColor}>
+            {children}
+          </Text>
+        </span>
+      )}
+    </RootComponent>
+  );
+}
```
OskiTheCoder, line 134: **question**: in the future if we intend to support animations, do you know how much, if it all, this component architecture would need to change? i guess it depends on which animation library we go with, but curious if theres any obvious we'd need to do.

OskiTheCoder:
**question**:  i noticed the text was getting highlighted when i clicked the toggle a few times, is this intentional or something coming from aria?


https://github.com/EasyPost/easy-ui/assets/44451175/a267cad8-1073-408f-90d4-fad5ffe6a65b

easy-ui-react/src/Toggle/Toggle.tsx
stephenjwatkins: it shouldn't be too difficult. Aria actually does a good job giving us the values in code for differentiating between states so that it could plug into a React animation library fairly easily

stephenjwatkins:
> **question**: i noticed the text was getting highlighted when i clicked the toggle a few times, is this intentional or something coming from aria?

good catch. i used the example from Aria which turns out has the same problem. then i went to their Spectrum component and noticed that it wasn't having that issue, so i copied the approach they use in Spectrum instead. should be fixed now

easy-ui-react/src/styles/_accessibility.scss
```diff
@@ -16,6 +16,25 @@
   white-space: nowrap !important;
 }
 
+/// Used to overlay a hidden input on top of a custom-styled control so that
+/// the input will receive user events while only showing the custom control.
+@mixin overlay-input {
```
OskiTheCoder, line 21: nice

easy-ui-react/src/Toggle/Switch.module.scss
```diff
@@ -0,0 +1,71 @@
...
+    theme-token("color.background.gray")
+  );
+  @include component-token(
+    "switch",
+    "color.thumb",
+    theme-token("color.text.primary")
+  );
+
+  display: inline-flex;
+  border-radius: 99px;
```
OskiTheCoder, line 16: **question**: could this be wrapped in a `component-token`?

PR #282, opened 2023-06-07
Revert "Bump vitest from 0.31.4 to 0.32.0"

stephenjwatkins:
Reverts EasyPost/easy-ui#281

Broke our tests since it's a breaking change.

We need to get the CI status check requirement in place.


stephenjwatkins:
going to just fix the problem instead

PR #292, opened 2023-06-08
Create `<RadioGroup />` component

stephenjwatkins:
## 📝 Changes

- Creates RadioGroup component
- Updates layout primitives to be more flexible
- Creates SelectorErrorTooltip to share across checkboxes and radios

## ✅ Checklist

- [x] Changeset is added


easy-ui-react/src/VerticalStack/VerticalStack.tsx
```diff
@@ -39,15 +36,24 @@ export type VerticalStackProps = {
   /** Vertical alignment of children */
   align?: Align;
 
-  /** Horizontal alignment of children */
-  inlineAlign?: InlineAlign;
+  /** Contents of the vertical stack. */
+  children: React.ReactNode;
+
+  /** Custom className for the vertical stack. */
+  className?: string;
```
OskiTheCoder, line 43: **question** so i see the use for this and its sensible, but do you have any concerns about having this exposed and this also being a component consumers can use?

easy-ui-react/src/HorizontalStack/HorizontalStack.tsx
```diff
@@ -83,10 +92,15 @@ export const HorizontalStack = React.forwardRef<null, HorizontalStackProps>(
...
+        inline ? "inline-flex" : "flex",
+      ),
     } as React.CSSProperties;
     return (
       <As
-        className={styles.HorizontalStack}
+        className={classNames(styles.HorizontalStack, className)}
         style={style}
         ref={ref}
         {...restProps}
```
OskiTheCoder, line 106: **question** it doesnt have to be part of this task,  but do you think at some point it might be worth adding some unit tests to the stack components?

documentation/specs/RadioGroup.md
```diff
@@ -42,9 +42,6 @@ type RadioGroupProps = {
   /** Whether the radio is disabled. */
   isDisabled?: boolean;
 
-  /** Whether or not the radio group label is visually hidden. Only recommended for when the grouping is visually implicit. */
-  isLabelVisuallyHidden?: boolean;
```
OskiTheCoder, line 46: **question** does it not make sense for this component to be able to visually hide a label?

easy-ui-react/src/HorizontalStack/HorizontalStack.tsx
stephenjwatkins: i think that's a good idea. i added a comment on the task for creating stories

easy-ui-react/src/VerticalStack/VerticalStack.tsx
stephenjwatkins: i do have concerns, yes. i think you're right. i removed it and modified the component structure

documentation/specs/RadioGroup.md
stephenjwatkins: this is a good question and became slightly more complicated when i implemented the component. yes it does make sense to allow for visually hiding the label. however react aria has a built-in mechanism for allowing the user to provide a custom aria label either through `aria-label`, `aria-labelledby`, `aria-describedby`, and/or `aria-details`. in fact, react aria writes a console warning describing what to do when a label is not passed. with that built-in mechanism, instead of using `isLabelVisuallyHidden` which would limit us to assigning an `aria-label` on the input, i thought it'd be best to support the react aria way of doing it, allowing for `label` to be optional, but in which case the consumer would need to pass one of the labeling props. then the consumer could use `aria-labelledby`, as an example, if they have a more complex use case beyond just an `aria-label`.

i actually added to this type the `AriaLabelingProps` from `react-aria` that supports the props mentioned above.

let me know if that makes sense

PR #308, opened 2023-06-13
Create `<CodeBlock />` spec

stephenjwatkins:
## 📝 Changes

- Create `<CodeBlock />` spec

### DIFF
--- a/documentation/specs/CodeBlock.md
+++ b/documentation/specs/CodeBlock.md
@@ -1,0 +1,204 @@
+# `CodeBlock` Component Specification
+
+## Overview
+
+A Code Block is used to display readable blocks of code.
+
+### Use Cases
+
+- Use a Code Block to improve readability of embedded code samples with syntax highlighting and automatic line numbering.
+
+### Features
+
+- Supports optional line numbering
+- Supports maximum number of lines before scroll
+- Supports multiple languages
+
+### Prior Art
+
+- [Paste `<CodeBlock />`](https://paste.twilio.design/components/code-block)
+- [Atlassian `<CodeBlock />`](https://atlassian.design/components/code/code-block)
+- [Sentry `<CodeSnippet />`](https://storybook.sentry.dev/?path=/story/components-code-snippet--js)
+
+---
+
+## Design
+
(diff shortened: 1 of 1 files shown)


documentation/specs/CodeBlock.md
```diff
@@ -0,0 +1,204 @@
...
+- [Atlassian `<CodeBlock />`](https://atlassian.design/components/code/code-block)
+- [Sentry `<CodeSnippet />`](https://storybook.sentry.dev/?path=/story/components-code-snippet--js)
+
+---
+
+## Design
+
+`CodeBlock` at a base level is pretty simple. It requires a `code` snippet and `language` for highlighting. It can optionally show line numbers and render a maximum number of lines before scrolling. It renders inside of a `Card`.
+
+`CodeBlock` will utilize the `PrismLight` export of `react-syntax-highlighter` to enable syntax highlighting. A lightweight wrapper component called `SyntaxHighlighter` will be created to wrap some details specific to EasyPost, including importing required languages and building some utilities around theme customization.
```
stephenjwatkins, line 29: uses Prism under the hood (a lighter version of it, which means you manage the import) but it seems to be a more robust setup compared to what we have used in easypost-web, at least from what i've used of it so far

documentation/specs/CodeBlock.md
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Constrains the height of code block to a set number of lines.
+   */
+  maxLines?: number;
+
+  /**
+   * Include line numbers in code block.
+   */
+  showLineNumbers?: boolean;
```
stephenjwatkins, line 78: i asked design regarding the alternative backbone style that's in figma (the carets) and we agreed to not worry about that for now since the syntax highlighting libs didn't offer a solution for that out of the box

documentation/specs/CodeBlock.md
```diff
@@ -0,0 +1,204 @@
...
+<CodeBlock.Container>
+  <CodeBlock.Header label="filename.js" />
+  <CodeBlock.Snippet code={`console.log("Hello world");`} />
+</CodeBlock.Container>
+```
+
+### API
+
+```ts
+export type SnippetLanguages =
```
stephenjwatkins, line 43: these are the languages i see our docs using

PR #311, opened 2023-06-14
select spec

OskiTheCoder:
## 📝 Changes

- adds `<Select />` spec

### DIFF
--- a/documentation/specs/Select.md
+++ b/documentation/specs/Select.md
@@ -1,0 +1,199 @@
+# `Select` Component Specification
+
+## Overview
+
+The `Select` component allows users to select a value from a set of options.
+
+### Use Cases
+
+- Can be used as a traditional select on forms.
+
+### Features
+
+- Setting the `size` property also sets size for iconAtStart: The size values map to Easy UI's token sizes for icons.
+- When `errorText` is supplied with `validationState="invalid"`, `helperText` is overriden.
+- `isLabelVisuallyHidden` property can be used to visually hide the label.
+- Supports disabled state.
+
+### Risks and Challenges
+
+- Ensuring dropdown doesn't inadvertently get cut off.
+- Maintaining an approachable and familiar API; similar to existing Easy UI form related components.
+- Ensuring the component remains accessible when label is visually hidden.
+
+### Prior Art
+
+- [Paste `<Select />`](https://paste.twilio.design/components/select)
+- [Polaris `<Select />`](https://polaris.shopify.com/components/selection-and-input/select)
+- [Spectrum `<Picker />`](https://react-spectrum.adobe.com/react-spectrum/Picker.html)
+
+---
+
+## Design
+
+### API
+
+```ts
+import type { AriaSelectProps } from "react-aria";
+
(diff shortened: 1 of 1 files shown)


documentation/specs/Select.md
```diff
@@ -0,0 +1,216 @@
...
+  /** The contents of the collection. */
+  children: CollectionChildren<T>;
+  /** Item objects in the collection. */
+  items?: Iterable<T>;
+  /** The item keys that are disabled. These items cannot be selected, focused, or otherwise interacted with. */
+  disabledKeys?: Iterable<Key>;
+};
+```
+
+### Anatomy
```
OskiTheCoder, line 106: my initial approach was to try to make use of the `Menu` in a way where i could get away with just defining my trigger, the select field itself. i ran into a few issues with that approach with regards to state management and aria props, and i felt the way i was needing to extend `Menu` felt unnecessarily awkward/complex. i had a look at how Spectrum approached this and it looks like they have a common `Popover` component. depending on how this implementation shakes out, it might be useful for us in the future to create some kind of generic popover. for now i think a decent path forward is to implement the `Select` by using the react-aria/react-stately select related hooks and by taking heavy inspiration from the existing `Menu` and how it handles the `OverlayContainer` and `Item`.

stephenjwatkins: that seems like a good path to me. once we have two implementations it may be easier to see how the two align

the only concern is ensuring we're not unnecessarily duplicating configuration that should change together. if the popup menu styles should be the same for the select dropdown and the menu, for instance, it would be nice to figure out how to share those configurations so they don't get out of sync

documentation/specs/Select.md
```diff
@@ -0,0 +1,216 @@
...
+  /** The currently selected key in the collection (controlled). */
+  selectedKey?: Key | null;
+  /** The initial selected key in the collection (uncontrolled). */
+  defaultSelectedKey?: Key;
+  /** Handler that is called when the selection changes. */
+  onSelectionChange?: (key: Key) => any;
+  /** The contents of the collection. */
+  children: CollectionChildren<T>;
+  /** Item objects in the collection. */
+  items?: Iterable<T>;
```
stephenjwatkins, line 100: **thought:** i feel iffy on having these two approaches for defining items. reading through the stately docs, i understand that using `items` gives you performance benefits, but i think it's confusing for consumers to have to decide when to use what.

i'm personally of the opinion that we should only expose one paradigm in our components based on what we feel makes the most sense for the component. there may be some, like a table, that needs to be concerned with performance and so we need to use `items`, but i think a select shouldn't need to. in which case, i would be a proponent of only having one option for adding items using the `children` prop.

so for dynamic data the consumer can use the same approach as they would with static data and map through the array elements:

```tsx
<ListBox>
  {animals.map(item =>
    <Item key={item.name}>{item.name}</Item>
  )}
</ListBox>
```

what do you think?

documentation/specs/Select.md
OskiTheCoder: agree

stephenjwatkins (approved):
lgtm!

PR #319, opened 2023-06-15
Version packages

stephenjwatkins:
## 📝 Changes

- Version packages to `-alpha.4`

### To publish a new version of Easy UI

- make sure you're on latest `main` (`git checkout main && git pull`)
- version the packages (`npm run changes:version`) (this processes all the pending changesets and generates CHANGELOG entries)
- look over the CHANGELOG entries. make sure dev-only packages aren't included (eslint, stylelint, tsconfig, etc). then commit the changes (`git add --all && git commit -m "Version packages"`)
- create a PR (like this one) with the versioned packages commit above. get approval of the PR from team
- merge the PR. re-pull `main`. then publish the packages (`git checkout main && git pull`; `npm run changes: publish`). make sure you have write access to NPM and ensure your registry isn't connected to Cloudsmith, otherwise it'll publish there instead
- `git push --follow-tags` to push the tags that were created in the publishing step

this is similar to what's documented here under "If you don't want to use this action, the manual workflow we recommend for running the version and publish commands": https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md#how-do-i-run-the-version-and-publish-commands

(i was going to write this in documentation somewhere but not sure right now where to put it so just leaving here for now)

### DIFF
--- a/.changeset/chilled-meals-whisper.md
+++ b/.changeset/chilled-meals-whisper.md
@@ -1,5 +1,4 @@
 ---
-"@easypost/stylelint-easy-ui": minor
 "@easypost/easy-ui-tokens": minor
 "@easypost/easy-ui": minor
 ---
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -10,15 +10,23 @@
     "@easypost/stylelint-easy-ui": "0.0.0"
   },
   "changesets": [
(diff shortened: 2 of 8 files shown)


OskiTheCoder (approved):
🎉

ldewald:
Published, this version is live

PR #327, opened 2023-06-21
create select component

OskiTheCoder:
## 📝 Changes

Adds `<Select />` component

## ✅ Checklist

- [x] Changeset is added


easy-ui-react/src/Select/SelectField.tsx
```diff
@@ -0,0 +1,133 @@
...
+        {...labelProps}
+      >
+        {label}
+      </Label>
+      <HiddenSelect
+        isDisabled={isDisabled}
+        state={selectState}
+        triggerRef={triggerRef}
+        label={label}
+      />
```
OskiTheCoder, line 109: react-aria's `HiddenSelect` component is causing an accessibility violation according to storybook:

sb_violation

according to react-aria, changing the `visibility` to `hidden` would break some functionality on safari: https://github.com/adobe/react-spectrum/blob/main/packages/@react-aria/select/src/HiddenSelect.tsx#L53

easy-ui-react/src/Select/SelectOverlay.tsx
```diff
@@ -0,0 +1,103 @@
+import React from "react";
```
OskiTheCoder, line 1: instead of having two overlays (one for menu and one for select), i think it would be good at some point to explore creating a generic `Popover` component that remains internal to Easy UI that supports both the select's listbox and the existing menu. i initially was hoping i could do it here but the overall PR was getting quite complex and i think its something that requires its own care/attention.

easy-ui-react/src/Select/Select.tsx
```diff
@@ -0,0 +1,166 @@
...
+import { SelectSection } from "./SelectSection";
+import { SelectOverlay } from "./SelectOverlay";
+import { useTriggerWidth } from "../Menu/useTriggerWidth";
+
+export type SelectProps<T> = AriaSelectProps<T> &
+  SelectFieldProps & {
+    /** Method that is called when the open state of the select field changes. */
+    onOpenChange?: (isOpen: boolean) => void;
+    /** Sets the open state of the select field. */
+    isOpen?: boolean;
```
OskiTheCoder, line 17: not sure if it makes sense for a `Select` to have these handlers; i could go either way on this

documentation/specs/Select.md
```diff
@@ -65,45 +83,36 @@ export type SelectProps = AriaSelectProps & {
...
+  errorTextProps?: DOMAttributes<FocusableElement>;
+  /** Field value props. */
+  valueProps?: DOMAttributes<FocusableElement>;
 };
 ```
 
 ### Anatomy
 
-The `Select` component's behavior and accessibility implementation will be handled by React Aria's `useSelect` hook. State managment will be handled by React Stately's `useSelectState` hook. To handle the associated dropdown menu, `OverlayContainer` and `usePopover` will be used from React Aria. The implementation for the menu will take heavy inspiration from Easy UI's `MenuOverlay`.
+The `Select` component's behavior and accessibility implementation will be handled by React Aria's `useSelect` hook. State managment will be handled by React Stately's `useSelectState` hook. To handle the associated listbox, `OverlayContainer`, `useListBox`, and `usePopover` will be used from React Aria. For the options, `useOption` will be used from React Aria. The `Select` component will also support sections and will use `useListBoxSection` and `useSeparator` from React Aria. The implementation for the listbox will take heavy inspiration from Easy UI's `MenuOverlay`.
```
OskiTheCoder, line 115: wording here feels a little awkward, open to ideas on rephrasing it

easy-ui-react/src/Select/SelectSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { separatorProps } = useSeparator({ elementType: "li" });
+  return (
+    <>
+      {section.key !== state.collection.getFirstKey() && (
+        <li {...separatorProps} className={styles.separator} />
+      )}
+      <li {...itemProps}>
+        {section.hasChildNodes && (
+          <ul {...groupProps} className={styles.sectionList}>
+            {[...section.childNodes].map((item) => {
```
OskiTheCoder, line 50: `childNodes` is deprecated, but i dont think react-aria has given an alternative yet that we could readily use here.

.changeset/shaggy-readers-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@easypost/easy-ui": minor
```
ldewald, line 2: **Question: ** Do we add changes for new icon files?

easy-ui-react/src/Select/SelectOverlay.tsx
ldewald: +1 and good call keeping it out of this PR haha

ldewald (approved):

.changeset/shaggy-readers-argue.md
OskiTheCoder: ah yes, youre right, updated PR

PR #373, opened 2023-07-03
docs for Button, IconButton, DropdownButton, and Icon

OskiTheCoder:
## 📝 Changes

- Adds basic tsdocs for `<Button />`, `<IconButton />`, `<DropdownButton />`, and `<Icon />`

### DIFF
--- a/easy-ui-react/src/Button/Button.tsx
+++ b/easy-ui-react/src/Button/Button.tsx
@@ -45,8 +45,8 @@ export type ButtonProps = AriaButtonProps & {
  * Allow users to perform an action or to navigate to another page.
  *
  * @remarks
- * Use to call attention to a place in the UI where the user can act in order
- * to continue flow. Supports icons, multiple colors, outlined and link variations,
+ * Use to communicate an action the user can take in order to continue
+ * flow in the UI. Supports icons, multiple colors, outlined and link variations,
  * and small size buttons. Underlying element is a button, but becomes an
  * anchor when `href` is provided.
  *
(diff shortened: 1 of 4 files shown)


easy-ui-react/src/Button/Button.tsx
```diff
@@ -41,6 +41,67 @@ export type ButtonProps = AriaButtonProps & {
   href?: string;
 };
 
+/**
+ * Allow users to preform an action or to navigate to another page.
```
ldewald, line 45: **Typo:** s/preform/perform
```suggestion
 * Allow users to perform an action or to navigate to another page.
```

easy-ui-react/src/Button/Button.tsx
```diff
@@ -41,6 +41,67 @@ export type ButtonProps = AriaButtonProps & {
...
+ *
+ * @remarks
+ * Use to call attention to a place in the UI where the user needs to do
```
ldewald, line 48: **Nit:** I'm not sure _needs_ is the right word since often it's by choice but I don't have an alternate suggestion

OskiTheCoder: updated sentence, open to other suggestions/ideas

easy-ui-react/src/Button/Button.tsx
OskiTheCoder: ah good catch

ldewald (approved):
Sorry for the delay, I thought I had already hit approve!

PR #406, opened 2023-07-12
Create `<CodeBlock />` spec

stephenjwatkins:
## 📝 Changes

- Adds CodeBlock spec

### DIFF
--- a/documentation/specs/CodeVisualizer.md
+++ b/documentation/specs/CodeVisualizer.md
@@ -1,17 +1,19 @@
-# `CodeVisualizer` Component Specification
+# `CodeBlock` Component Specification
 
 ## Design
 
-`CodeVisualizer` wraps one or more `CodeBlock`s with a header to allow the user to switch between blocks.
+**Our current `CodeBlock` component will be renamed to `CodeSnippet`.**
 
-`CodeVisualizer` will be a controlled component. It will require the consumer to provide a selected language and manage its state.
+`CodeBlock` wraps one or more `CodeSnippet`s with a header to allow the user to switch between them.
+
+`CodeBlock` will be a controlled component. It will require the consumer to provide a selected language and manage its state.
 
 ### API
 
 ```ts
-export type CodeVisualizerProps = {
+type CodeBlockProps = {
   /**
-   * CodeVisualizer content. This should be a header and collection of snippets.
+   * CodeBlock content. This should be a header and collection of snippets.
    */
   children: ReactNode;
 
@@ -26,7 +28,7 @@ export type CodeVisualizerProps = {
   onLanguageChange: (language: SnippetLanguages) => void;
 };
 
-export type CodeVisualizerHeaderProps = {
+type CodeBlockHeaderProps = {
   /**
    * Header title.
    */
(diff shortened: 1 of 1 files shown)


documentation/specs/CodeVisualizer.md
```diff
@@ -0,0 +1,135 @@
+# `CodeVisualizer` Component Specification
+
+## Design
```
stephenjwatkins, line 3: the design for this was heavily inspired by Twilio's: https://paste.twilio.design/components/code-block#tabs

documentation/specs/CodeVisualizer.md
```diff
@@ -0,0 +1,135 @@
...
+
+`CodeVisualizer` wraps one or more `CodeBlock`s with a header to allow the user to switch between blocks.
+
+`CodeVisualizer` will be a controlled component. It will require the consumer to provide a selected language and manage its state.
```
stephenjwatkins, line 7: how do we feel about it being only controlled? only controlled simplifies our implementation but may be a little more friction for the consumer. alternatively we could implement for it allowing both controlled and uncontrolled at the cost of a little more implementation complexity

documentation/specs/CodeVisualizer.md
```diff
@@ -0,0 +1,135 @@
...
+
+### API
+
+```ts
+export type CodeVisualizerProps = {
+  /**
+   * CodeVisualizer content. This should be a header and collection of snippets.
+   */
+  children: ReactNode;
```
stephenjwatkins, line 16: if there's only one snippet provided, the CodeVisualizer goes into "View Only" mode and doesn't show a language dropdown

documentation/specs/CodeVisualizer.md
```diff
@@ -0,0 +1,135 @@
...
+   * Code snippet content. This should contain a CodeBlock.
+   */
+  children: ReactNode;
+};
+```
+
+### Example Usage
+
+```tsx
+import { CodeVisualizer } from "@easypost/easy-ui/CodeBlock";
```
OskiTheCoder, line 54: **question**: is the language picker/selector intended to function as a trigger for a menu/listbox? if so, do you have a rough idea for how you want to implement the trigger?

documentation/specs/CodeVisualizer.md
```diff
@@ -0,0 +1,135 @@
+# `CodeVisualizer` Component Specification
```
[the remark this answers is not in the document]
OskiTheCoder: im ultimately ok with `CodeVisualizer`, but just some thoughts: at first glance when i see `CodeBlock` and `CodeVisualizer`, its not immediately obvious to me which is the _molecule_ level component. `CodeBlockWrapper` feels a bit more descriptive and ive seen it out in the wild too, but it also feels kinda verbose to say/write lol.

documentation/specs/CodeVisualizer.md
OskiTheCoder: im personally a fan of it only being controlled based on how im understanding the use case

documentation/specs/CodeVisualizer.md
stephenjwatkins: yeah it's supposed to trigger our `Menu`. i was imagining using an `UnstyledButton` with a `Text` and `Icon` inside as the `Menu.Trigger`

documentation/specs/CodeVisualizer.md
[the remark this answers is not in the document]
ldewald: I agree it's not obvious which is the molecule level. What if we renamed the existing `CodeBlock` to be `CodeHighlighter` (or `SyntaxHighlighter`) and used `CodeBlock` for this one? I'm okay with `CodeVisualizer` as well but throwing this idea out there.

documentation/specs/CodeVisualizer.md
ldewald: > im personally a fan of it only being controlled based on how im understanding the use case

+1

documentation/specs/CodeVisualizer.md
```diff
@@ -0,0 +1,135 @@
...
+
+```tsx
+import { CodeVisualizer } from "@easypost/easy-ui/CodeBlock";
+
+function Component() {
+  const [language, setLanguage] = useState("javascript");
+  return (
+    <CodeVisualizer language={language} onLanguageChange={setLanguage}>
+      <CodeVisualizer.Header>Header</CodeVisualizer.Header>
+      <CodeVisualizer.Snippet>
```
ldewald, line 61: Curious what this wrapper is for, is there a case where we what anything other than (or in addition to) `CodeBlock`?

PR #462, opened 2023-07-27
Add `<Modal />` spec

stephenjwatkins:
## 📝 Changes

- Adds modal spec

### DIFF
--- a/documentation/specs/Modal.md
+++ b/documentation/specs/Modal.md
@@ -1,0 +1,162 @@
+# `Modal` Component Specification
+
+## Overview
+
+A `Modal` is a page overlay that displays information and blocks interaction with the page until an action is taken or the `Modal` is dismissed.
+
+## Design
+
+`Modal` will use `useDialog` and `useModalOverlay` from `react-aria` to provide the technical foundation for an accessible dialog.
+
+The component design was inspired both by Shopify's Modal component and Twilio's Modal component.
+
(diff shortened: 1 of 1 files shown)


documentation/specs/Modal.md
```diff
@@ -0,0 +1,191 @@
...
+   * Callback when the modal is closed.
+   */
+  onClose: () => void;
+
+  /**
+   * Whether or not the modal can be dismissed.
+   *
+   * @default true
+   */
+  isDismissable?: boolean;
```
stephenjwatkins, line 38: i don't love having a boolean prop that defaults to `true`. i rather the API require the boolean to be active to make it read better in the tree. however i couldn't think of a clean name that wasn't a negation like `isNotDismissable` which read less clean IMO

documentation/specs/Modal.md
```diff
@@ -0,0 +1,191 @@
...
+   */
+  subtitle?: string;
+
+  /**
+   * Icon to display at the start of the header title.
+   */
+  iconAtStart?: {
+    accessibilityLabel?: string;
+    symbol: IconSymbol;
+  };
```
stephenjwatkins, line 65: i don't love having complex props like this but i couldn't think of a cleaner way to handle these slots that need to be flexible but only to an extent. inspired by Polaris' Modal who does similar for some of the slots. i'm open to better ideas.

OskiTheCoder (approved):
lgtm, exciting stuff!

PR #478, opened 2023-08-01
Add `<Modal />` component

stephenjwatkins:
## 📝 Changes

- Adds `<Modal />` component
- Fix `<Icon />` css for centering smaller symbols
- Fix `<HorizontalStack />` css for child truncation

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added


easy-ui-react/src/Icon/Icon.module.scss
```diff
@@ -7,10 +7,12 @@
     height: $var;
   }
 
-  display: block;
+  display: flex;
```
stephenjwatkins, line 10: fixes some cases where the icon will incorrectly mess up the parent container's layout behavior

easy-ui-react/src/Modal/useIntersectionDetection.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!targetRef.current || !scrollRef.current) {
+      return;
+    }
+
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        setIsStuck(!entry.isIntersecting);
+      },
+      { root: scrollRef.current },
+    );
```
OskiTheCoder, line 26: **question**: hmmm im probably missing something obvious, but what is this bit doing?

`{ root: scrollRef.current }`

easy-ui-react/src/Modal/Modal.module.scss
```diff
@@ -0,0 +1,139 @@
+@use "../styles/common" as *;
+@use "../styles/unstyled";
+
+.underlayBg {
+  @include component-token("modal", "underlay-edge-max-height", 72px);
+  @include component-token("modal", "underlay-edge-min-height", 2.5vh);
```
OskiTheCoder, line 6: **question**: how was the min height here calculated?

easy-ui-react/src/Icon/Icon.module.scss
OskiTheCoder: i had been seeing some folks using an additional wrapper around icons, this addresses that, nice!

easy-ui-react/src/Modal/Modal.module.scss
stephenjwatkins: it's computed from the requirement in Figma for the max height of the modal to be "95% of the screen height" which leaves each top/bottom edge 2.5% or 2.5vh. let me know if that makes sense.

image

easy-ui-react/src/Modal/useIntersectionDetection.ts
stephenjwatkins: it's passing in a custom scroll root to the intersection observer so that the intersection can be based off of a non-body scroll container, in the case of our modals the scroll root is the body of the Modal. does that make sense? i can maybe update the naming to reflect that better

OskiTheCoder: ahh! i see, yes that makes sense. `scrollRef` seems quite appropriate then, im good with the naming

easy-ui-react/src/Modal/Modal.module.scss
OskiTheCoder: makes sense!

OskiTheCoder (approved):
lgtm! from what i can tell covers all the intended use cases, nice work!

PR #488, opened 2023-08-03
fix(TextField,Textarea): pass all properties through to InputField

stephenjwatkins:
## 📝 Changes

- Passes properties through to `InputField`. Fixes problem of consumers not having their events registered
- Adds tests for use case

## ✅ Checklist

- [x] Changeset is added


OskiTheCoder (approved):
nice, thanks for fixing this!

PR #524, opened 2023-08-11
Create `<DataGrid />` spec

stephenjwatkins:
## 📝 Changes

- Creates `<DataGrid />` spec


documentation/specs/Table.md
```diff
@@ -0,0 +1,354 @@
...
+  defaultExpandedKey?: Key;
+
+  /** The initial selected keys in the collection (uncontrolled). */
+  defaultSelectedKeys?: "all" | Iterable<Key>;
+
+  /** The currently expanded key in the collection (controlled). */
+  expandedKey?: Key;
+
+  /** Whether or not row expansion is enabled. */
+  hasExpandableRows?: boolean;
```
ldewald, line 56: **Question:** Are you thinking there's a case where we would provide any of the other expandable props (`expandedKey`, `onExpandedChange`, `renderExpandedRow` but want to disable expandability?

stephenjwatkins: no i don't imagine that being a case. and i went back and forth on this option as a mechanism for expressed intentionality of row expansion, but i kinda tend to think it's superfluous with the other props

ldewald (approved):
Looks great

PR #535, opened 2023-08-15
Create `<DisplayTable />` spec

stephenjwatkins:
## 📝 Changes

- Creates `<DisplayTable />` spec

### DIFF
--- a/documentation/specs/DisplayTable.md
+++ b/documentation/specs/DisplayTable.md
@@ -1,0 +1,31 @@
+# `DisplayTable` Component Specification
+
+## Overview
+
+A `DisplayTable` is used for presenting information displayed across columns and rows.
+
+---
+
+## Design
+
+`DisplayTable` is only for presentational concerns. It doesn't support advanced capabilities such as selection, sorting, and row expansion. With that in mind, complexities are primarily shifted to surfacing a simple component API, structuring the table properly to support the style requirements, and ensuring that accessibility markup is applied correctly.
+
+Even though `DisplayTable` doesn't do much in terms of functionality, because accessibility concerns can be non-trivial, we will use `useTable` from React Aria to ensure that we have our accessibility bases covered.
+
+### API
+
+```ts
+
+```
+
+### Example Usage
+
+_Simple_:
+
+```tsx
+import { Modal } from "@easypost/easy-ui/Modal";
+
+function PageWithModal() {
+  return <DisplayTable />;
+}
+```
--- a/documentation/specs/DisplayTable.md
+++ b/documentation/specs/DisplayTable.md
@@ -8,24 +8,221 @@ A `DisplayTable` is used for presenting information displayed across columns and
 
 ## Design
 
-`DisplayTable` is only for presentational concerns. It doesn't support advanced capabilities such as selection, sorting, and row expansion. With that in mind, complexities are primarily shifted to surfacing a simple component API, structuring the table properly to support the style requirements, and ensuring that accessibility markup is applied correctly.
+`DisplayTable` is for presentational concerns only. It doesn't support interaction such as selection, sorting, and row expansion. For an interactive table accommodating advanced use cases, see `Table`.
 
-Even though `DisplayTable` doesn't do much in terms of functionality, because accessibility concerns can be non-trivial, we will use `useTable` from React Aria to ensure that we have our accessibility bases covered.
+Without interaction, complexities are shifted to surfacing a simple component API, structuring the table for style requirements, and ensuring appropriate application of accessibility markup.
+
+`DisplayTable` will use React Aria's `useTable` under the hood to reliably generate accessibility markup for the table.
 
 ### API
 
 ```ts
+type DisplayTableProps = AriaLabelingProps & {
+  /** Columns for the table. */
+  columns: Column[];
 
+  /**
+   * Horizontal alignment of data within their cells.
+   * @default start
+   */
+  dataAlignment?: "start" | "center" | "end";
+
+  /** Renders the first row cell as a row header. */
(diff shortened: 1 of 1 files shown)


documentation/specs/DisplayTable.md
```diff
@@ -0,0 +1,238 @@
...
+  /**
+   * Horizontal alignment of data within their cells.
+   * @default start
+   */
+  dataAlignment?: "start" | "center" | "end";
+
+  /**
+   * Marks the table has having no row headers. For accessibility reasons, this should be avoided when possible.
+   */
+  hasNoRowHeaders?: boolean;
```
stephenjwatkins, line 35: instead of checking for a row header through a `heading` property in the first cell of the row data, control it as a prop in the `DisplayTable` component, since it should be table-wide setting instead of a cell-specific setting (either all or no rows should have row headers).

name is negated on purpose to discourage not having row headers. open to feedback on a better name.

documentation/specs/DisplayTable.md
```diff
@@ -0,0 +1,238 @@
...
+type Row = Cell[];
+
+type Cell =
+  | ReactNode
+  | {
+      /** Cell text. */
+      text?: ReactNode;
+
+      /** Cell icon. */
+      symbol?: IconSymbol;
```
stephenjwatkins, line 50: this will likely be the hardest part in migrating www to the new table. using our new Icon system instead of using webfont glyphs. likely we can take advantage of codemods to undertake this

documentation/specs/DisplayTable.md
```diff
@@ -0,0 +1,238 @@
+# `DisplayTable` Component Specification
```
stephenjwatkins, line 1: tried to name this to be agnostic of its location. in theory it could be used in our products and so i didn't want the name to limit us in that regard. open to better names

documentation/specs/DisplayTable.md
```diff
@@ -0,0 +1,238 @@
...
+`DisplayTable` will use React Aria's `useTable` under the hood to reliably generate accessibile markup for the table.
+
+`DisplayTable` will use `columns` and `rows` as the primary mechanism for passing data to the table. While this is a slight departure from our dot-notation pattern used throughout the rest of Easy UI, this makes for a consistent data input pattern with our `Table`-related components.
+
+### API
+
+```ts
+type DisplayTableProps = AriaLabelingProps & {
+  /** Columns for the table. */
+  columns: Column[];
```
stephenjwatkins, line 24: for `DisplayTable`, `columns` and `rows` follow a highly-specific data structure as noted in the type definitions below. this is a bit of a departure from our `Table` component, where the actual cell data is largely arbitrary and the rendering is controlled by the consumer (via `renderHeaderCell`, `renderBodyCell`, etc).

i think (a) i wanted to try to match our current MarketingTable API as much as we could where reasonable; and (b) it seems like to me for a display table where so much of the styling is governed by the internals of the component itself, it makes sense for it to not to defer rendering to the consumer and instead opt for a specific data structure.

let me know if that makes sense and feels cohesive enough in our component library for having the two approaches, and/or if there are better ideas

documentation/specs/DisplayTable.md
ldewald: Looked at a few different systems for inspiration and seeing `Data Grid` come up as a name for what we are naming `Table` ("interactive tabular interface") and `Table` for this component ("display simple datasets that do not need to be filtered or edited"). Don't feel super strongly but I do like Data Grid for the other table we're building.
Goldman Sachs: [Data Grid](https://design.gs.com/components/data-grid)/[Table](https://design.gs.com/components/table)
Edison™ Design System: [Data Grid](https://edisondesignsystem.com/datagrid)/[Table](https://edisondesignsystem.com/table?tab=0)

documentation/specs/DisplayTable.md
```diff
@@ -0,0 +1,238 @@
...
+
+  /**
+   * Horizontal alignment of data within their cells.
+   * @default start
+   */
+  dataAlignment?: "start" | "center" | "end";
```
ldewald, line 30: **Thought:** In a future task, perhaps we could try to create a pattern for naming for our alignment props for a simpler DX. I've been finding myself having to reference the documentation multiple times to get the right prop name when using the stack components. We have a few that are used for various purposes (`align`, `alignInline`, `alignBlock`, `alignItems`) but thinking there could be an opportunity to simplify as `halign`/`valign` or similar.



align | Horizontal alignment of children
-- | --




align | Horizontal alignment of children
-- | --

documentation/specs/DisplayTable.md
ldewald: I think that would be a bonus to do, marketing was onboard with using unicode ✓ in place of the icon.

documentation/specs/DisplayTable.md
ldewald: I think this makes sense! One of the examples uses `hasRowHeaders` though ;)

documentation/specs/DisplayTable.md
stephenjwatkins: i agree. made a task to followup

documentation/specs/DisplayTable.md
stephenjwatkins: i like that. we'll use that scheme

documentation/specs/DisplayTable.md
stephenjwatkins: whoops good catch

ldewald (approved):

kevinalexliu:
This looks great!

kevinalexliu (approved):

PR #544, opened 2023-08-17
Write ADR for Page Layouts

stephenjwatkins:
## 📝 Changes

- Adds ADR for page layouts


documentation/decisions/012_page_layouts.md
```diff
@@ -0,0 +1,66 @@
...
+
+These components will include essential properties and hooks, enabling consuming sites to meet their individual design and feature requisites. This decision benefits our entire platform by promoting uniformity and upholding higher quality standards, particularly for page layouts being some of the more UI-challenging aspects of our sites.
+
+**Regarding Terminology:** Although we will incorporate these components into Easy UI, we intend to preserve the distinct meanings of "page" and "layout" found in web frameworks. "Pages" often represent terminal nodes in routing structures, whereas "layouts" signify branches. Both pages and layouts commonly involve specific business logic pertaining to their respective parts of the routing tree. Since Easy UI exclusively deals with presentation-related concerns, we will avoid naming components in a way that might confuse or overlap with these established semantics. Instead, we will adopt the term "Shell" (i.e. `ProductShell`), to explicitly indicate that it encompasses the outer presentation container of the corresponding layout.
```
stephenjwatkins, line 47: per usual, while i tend to believe in this sentiment, i'm not married to the term `Shell`, and am open to suggestions. from what i've seen of component libraries, they often use the suffix `Layout`, but from experience in our own codebase, that term easily becomes conflated and confusing, but happy to use it if it simply makes the most sense

ldewald: `Template` is another one to throw in the "what does this actually mean" ring. Shell makes sense to me and we can always change it later if folks find it confusing.

ldewald (approved):

documentation/decisions/012_page_layouts.md
stephenjwatkins: sounds good. agree that `Template` is another one to throw into the ring

PR #550, opened 2023-08-18
fix(Menu): return focus to trigger after closing menu

stephenjwatkins:
## 📝 Changes

- uses `Overlay` instead of `OverlayContainer` to return focus back to trigger after closing menu. this is per Aria documentation and was simply missed in the implementation
- attempted to write a unit test for this but couldn't get user event, jsdom, and Aria to play nicely together for this change. will come back to it when i figure out how to test focus in these cases

## ✅ Checklist

- [x] Changeset is added

### DIFF
--- a/.changeset/silent-beans-try.md
+++ b/.changeset/silent-beans-try.md
@@ -1,0 +1,5 @@
+---
+"@easypost/easy-ui": patch
+---
+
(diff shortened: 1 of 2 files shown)


ldewald (approved):
nice!

PR #569, opened 2023-08-24
Create `<DataGrid />` component

stephenjwatkins:
## 📝 Changes

- Adds `<DataGrid />` component
- Adds `standalone` support to `<Checkbox />`
- Preserve any custom props passed to `<Menu.Trigger />`

## ✅ Checklist

- [x] Changeset is added


easy-ui-react/src/Checkbox/Checkbox.test.tsx
```diff
@@ -95,8 +95,5 @@ describe("<Checkbox />", () => {
 });
 
 export async function selectCheckbox(user: UserEvent, el: HTMLElement) {
-  await act(async () => {
-    await user.click(el);
-    vi.runAllTimers();
-  });
+  await user.click(el);
```
stephenjwatkins, line 98: `vitest` appears to have fixed needing this wrapped in `act`

easy-ui-react/src/DataGrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,229 @@
...
+  { key: "permissions", name: "Permissions" },
+  { key: "status", name: "Status" },
+  { key: "lastActive", name: "Last Active" },
+];
+
+const rows = [
+  {
+    key: 1,
+    name: "Julie Smith",
+    email: "julie.smith@easypost.com",
```
ldewald, line 27: Consider using example.com instead in the (unlikely) event we get a Julie Smith or Sam Frost working here some day

easy-ui-react/src/DataGrid/useExpandedRow.ts
```diff
@@ -0,0 +1,177 @@
...
+  useMemo,
+  useState,
+} from "react";
+import { Node, TableState } from "react-stately";
+import { getComponentToken } from "../utilities/css";
+import { EXPAND_COLUMN_KEY } from "./constants";
+
+/**
+ * Manages the size and positioning of the expanded row content with
+ * minimal-to-no jank. It uses an invisible placeholder box to set and get the
```
ldewald, line 15: > minimal-to-no jank
😆

ldewald (approved):
I left a few non-blocking comments. I wouldn't have guessed the design required this level of complexity but the code looks great, nice work!

easy-ui-react/src/DataGrid/ColumnHeader.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const ColumnHeaderContentComponent = column.props.isSelectionCell
+    ? SelectAllColumnHeaderContent
+    : DefaultColumnHeaderContent;
+
+  return (
+    <div
+      ref={ref}
+      {...mergeProps(columnHeaderProps, focusProps)}
+      className={className}
+      data-ezui-data-grid-column-header="true"
```
OskiTheCoder, line 71: **question** is this property for ease of querying?

easy-ui-react/src/DataGrid/DataGrid.module.scss
```diff
@@ -0,0 +1,97 @@
...
+  border-radius: design-token("shape.border_radius.md");
+  /* stylelint-disable scss/operator-no-newline-after */
+  max-height: calc(
+    (component-token("data-grid", "border-width") * 2) +
+      component-token("data-grid", "header-row-height") +
+      (
+        component-token("data-grid", "body-row-height") *
+          component-token("data-grid", "max-rows")
+      )
+  );
```
OskiTheCoder, line 63: **question**: i think i get what this `max-height` is computing, but do we need it to prevent "height" overflow when expanding rows? or some other reason?

easy-ui-react/src/DataGrid/Table.tsx
```diff
@@ -0,0 +1,162 @@
...
+                </ExpandedRowContent>
+              )}
+            </Fragment>
+          ))}
+          {renderInterceptors()}
+        </div>
+      </div>
+    </DataGridTableContext.Provider>
+  );
+}
```
OskiTheCoder, line 162: **praise**: this table component is designed very well

easy-ui-react/src/DataGrid/useExpandedRow.ts
```diff
@@ -0,0 +1,177 @@
...
+ * appropriate bounding client rectangle representative of the expanded row
+ * content box, before applying it to the visible expanded row.
+ */
```
OskiTheCoder, line 18: **question**: so this computes the invisible box before actually applying it to the expanded row? this is probably a silly question and i might be missing something, but why is that necessary?

documentation/specs/DataGrid.md
```diff
@@ -96,6 +105,14 @@ type DataGridProps<C extends Column> = AriaLabelingProps & {
 
   /** The current sorted column and direction. */
   sortDescriptor?: SortDescriptor;
+
+  /**
+   * Define a custom grid-template-columns.
+   * Defaults to taking up an even amount of space.
+   *
+   * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns}
```
ldewald, line 113: This is rendering in Storybook with `}` as part of the link which is a 404
image

easy-ui-react/src/DataGrid/useGridTemplate.ts
```diff
@@ -0,0 +1,67 @@
...
+    pDefs.push("min-content");
+  }
+
+  if (columns.some((c) => c.key === ACTIONS_COLUMN_KEY)) {
+    sAreas.push("actions");
+    sDefs.push("min-content");
+  }
+
+  const areas = [pAreas, uAreas, sAreas]
+    .map((a) => a.join(" "))
```
vmike245, line 52: This can be changed to `.flat()`, which would remove the nesting of the array values so that they can be joined on the next line

easy-ui-react/src/DataGrid/useGridTemplate.ts
```diff
@@ -0,0 +1,67 @@
...
+    .join(" ")
+    .trim();
+
+  const defs = [pDefs, uDefs, sDefs]
+    .map((d) => d.join(" "))
```
vmike245, line 57: This can be changed to `.flat()`, which would remove the nesting of the array values so that they can be joined on the next line

easy-ui-react/src/DataGrid/ActionsCellContent.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }, [removeHover]);
+  return (
+    <Menu>
+      <Menu.Trigger>
+        <UnstyledPressButton onPress={handleClick}>
+          <Text visuallyHidden>{accessibilityLabel}</Text>
+          <Icon symbol={MoreVertIcon} />
+        </UnstyledPressButton>
+      </Menu.Trigger>
+      {rowAction.renderMenuOverlay()}
```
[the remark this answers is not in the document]
stephenjwatkins: i understand the tension here. i originally had a design like this, in fact, but decided on deferring the entire overlay so that the consumer can have direct access to the props of the `Menu.Overlay` component (`disabledKeys`, `maxItemsUntilScroll`, `onAction`, `onClose`, `placement`, and `width`) without `DataGrid` having to take care of plumbing in the implementation

easy-ui-react/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,222 @@
...
+      : defaultExpandedKey
+      ? defaultExpandedKey
+      : null;
+  });
+
+  const toggleExpandedRow = useCallback(
+    (rowKey: Key) => {
+      onExpandedChange(rowKey);
+      if (!isRowExpansionControlled) {
+        setExpandedKey((prevKey) => (prevKey === rowKey ? null : rowKey));
```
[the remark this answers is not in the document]
stephenjwatkins: `expandedKey` is a `Key`, which can be a `string` or `number` (or `null`). `setExpandedKey` is a `React.Dispatch<SetStateAction>` which allows for accepting a function with a previous value and returning the new `Key`, which is what's happening here

easy-ui-react/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,222 @@
...
+  const hasRowActions = Boolean(rowActions);
+  return useMemo(() => {
+    let c = columns;
+    if (hasExpandableRows) {
+      c = [{ key: EXPAND_COLUMN_KEY } as C, ...c];
+    }
+    if (hasRowActions) {
+      c = [...c, { key: ACTIONS_COLUMN_KEY } as C];
+    }
+    return c;
```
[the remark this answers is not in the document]
stephenjwatkins: early returns wouldn't allow for both conditions to process unfortunately

easy-ui-react/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,222 @@
...
+  return useMemo(() => {
+    const mappedRows = rows.map((row) => {
+      let r = row;
+      if (hasExpandableRows) {
+        r = { [EXPAND_COLUMN_KEY]: expandedKey === row.key, ...r };
+      }
+      if (hasRowActions) {
+        r = { ...r, [ACTIONS_COLUMN_KEY]: true };
+      }
+      return r;
```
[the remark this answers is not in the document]
stephenjwatkins: same as above

easy-ui-react/src/DataGrid/DataGrid.mdx
```diff
@@ -0,0 +1,235 @@
...
+];
+
+<DataGrid
+  aria-label="Example data grid"
+  columns={columns}
+  rows={rows}
+  renderColumnCell={(column) => <span>{String(column.name)}</span>}
+  renderRowCell={(item, columnKey, row) => <span>{String(item)}</span>}
+  // Enable row expansion
+  renderExpandedRow={(rowKey: Key) => <div>Expanded row content</div>}
```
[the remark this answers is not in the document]
stephenjwatkins: it was. thanks for the heads up. fixed

easy-ui-react/src/DataGrid/ColumnHeader.tsx
stephenjwatkins: yeah it is. which i don't love but i thought it'd be a little simpler than trying to manage refs throughout the DataGrid

easy-ui-react/src/DataGrid/DataGrid.module.scss
stephenjwatkins: yeah it's for setting an overflow on the table when a consumer sets `maxRows={}` to a certain number. that way if there are a lot of rows in the table the consumer can have it scroll past <x> number of rows instead of being super long. by default though this is disabled

documentation/specs/DataGrid.md
stephenjwatkins: i fixed the 404 by giving the URL a label, which is still valid JSDoc. unfortunate that Storybook doesn't parse it correctly

easy-ui-react/src/DataGrid/SortIndicator.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <span
+      aria-hidden="true"
+      className={classNames(
+        styles.SortIndicator,
+        isColumnSorted && styles.sorted,
+        headerVariant && styles[variationName("onHeader", headerVariant)],
+      )}
+    >
+      {isColumnSorted && sortDirection === "ascending" ? <Asc /> : <Desc />}
```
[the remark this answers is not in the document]
stephenjwatkins: i agree. updated to consolidate the logic into `sortDirection`

easy-ui-react/src/DataGrid/DataGrid.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  const isRowExpansionControlled = expandedKeyFromUser !== undefined;
+
+  const [expandedKey, setExpandedKey] = useState(() => {
+    return isRowExpansionControlled
+      ? expandedKeyFromUser
+      : defaultExpandedKey
+      ? defaultExpandedKey
+      : null;
+  });
```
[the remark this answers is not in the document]
stephenjwatkins: i love nested ternaries 🙃 but i updated it

easy-ui-react/src/DataGrid/DataGrid.stories.tsx
```diff
@@ -0,0 +1,229 @@
...
+    headerVariant: "secondary",
+  },
+  parameters: {
+    controls: {
+      include: ["headerVariant"],
+    },
+  },
+};
+
+export const WithSort: Story = {
```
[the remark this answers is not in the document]
stephenjwatkins: looks like it didn't like me referencing the `Template` from above. instead i referenced the `DataGrid` directly, which seems to fix it

stephenjwatkins:
> I wouldn't have guessed the design required this level of complexity

it surprised me a bit too. most of it was from needing to support row expansion, which wasn't directly supported by Aria out of the box

easy-ui-react/src/DataGrid/useExpandedRow.ts
stephenjwatkins: your comment made me realize that we don't need the invisible box after all. after creating this implementation, i refactored the DataGrid to use a separate scroll container from the grid itself. now we can update the box in one render cycle, before it required two which needed a second box to manage that render cycle delay. thanks lol!

OskiTheCoder (approved):
LGTM! its quite complex, but so are the requirements lol

easy-ui-react/src/DataGrid/DataGrid.tsx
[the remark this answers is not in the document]
vmike245: Oh yeah duh, I didn't even think about the fact that something could have both of those be true. No need to implement this change, but I tend to prefer optionally including items in an array instead of mutating it. So something like this:
```
return [
  ...(hasExpandableRows ? [{ key: EXPAND_COLUMN_KEY } as C] : []),
  ...columns,
  ...(hasRowActions ? [{ key: ACTIONS_COLUMN_KEY } as C] : []),
]
```
The reason that I prefer this format is because it indicates more clearly the intention of the code. The code that you have right now reads like this (at least to me): Create a new array from the value of columns, if it `hasExpandableRows` then create a new array that starts with `{ key: EXPAND_COLUMN_KEY }` and contains the rest of the old array and reassign the value of the variable that was just created, if it `hasRowActions` then create a new array that starts with `{ key: ACTIONS_COLUMN_KEY }` and contains the rest of the old array and reassign the value of the variable that was just created, finally return the newly created array.

Alternatively, my proposed code reads like: create and return a new array, set the first item in the returned array as `{ key: EXPAND_COLUMN_KEY }` if `hasExpandableRows` is true, include the items from columns in this array, set the last item in the returned array as `{ key: ACTIONS_COLUMN_KEY }` if `hasRowActions` is true.

Besides reading a little more clearly, it also minimizes places where things can go wrong. For instance when we do something like what is currently in the code we risk `if (hasRowActions)` breaking something that was done earlier (say we forget to spread `c` inside of the `if` statement). 

Like I said, no need to change this. I just wanted to point out a way that I think is a bit easier to read/understand in order to do something like this.

vmike245 (approved):

PR #621, opened 2023-09-15
Create `<TabNav />` spec

stephenjwatkins:
## 📝 Changes

- Creates `TabNav` spec


documentation/specs/TabNav.md
```diff
@@ -0,0 +1,81 @@
+# `TabNav` Component Specification
```
stephenjwatkins, line 1: since the name `Tabs` more often refers to the W3C Tabs ARIA pattern, which is not meant for navigation, i thought the name `TabNav`, which is what [Github's Primer](https://primer.style/design/components/tab-nav) uses for their version of tabs that navigate, would be more accurate of a name. per usual, i don't necessarily _love_ it. i've seen various different names in different component libraries for this component. open to suggestions

documentation/specs/TabNav.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Overview
+
+A `TabNav` is a set of styled links that let users navigate between related pages.
+
+---
+
+## Design
```
stephenjwatkins, line 9: note that the horizontal overflow behavior of this component (the responsive design in Figma) is still in flux. due to WAI ARIA reflow guidelines, that behavior may change to render overflowed items into a disclosure menu instead of using scroll. however the overflow is handled shouldn't cause a change in the API design of the component.

documentation/specs/TabNav.md
ldewald: Sounds good to me!

ldewald (approved):
Looks good to me

documentation/specs/TabNav.md
OskiTheCoder: agree, this is a sensible choice

OskiTheCoder (approved):
LGTM!

kevinalexliu (approved):
Looks good!!

PR #632, opened 2023-09-19
Create `<TabPanels />` spec

stephenjwatkins:
## 📝 Changes

- Adds `<TabPanels />` spec

### DIFF
--- a/documentation/specs/TabPanels.md
+++ b/documentation/specs/TabPanels.md
@@ -1,0 +1,163 @@
+# `TabPanels` Component Specification
+
+## Overview
+
+A `TabPanels` is a set of styled links that let users switch between multiple views within a related context under the same URL.
+
+Tabs in a tab panel should not be links that navigate the user to a new URL. If you want tab-like navigation, use `TabNav` instead.
+
+### Features
+
+- Support for mouse, touch, and keyboard interactions on tabs
+- Support for LTR and RTL keyboard navigation
+- Support for disabled tabs
+- Follows the tabs ARIA pattern, semantically linking tabs and their associated tab panels
+- Focus management for tab panels without any focusable children
+
+---
+
+## Design
+
+`TabPanels` uses React Aria's `useTabsList` under the hood. `useTabsList` provides the behavior and accessibility for a tab list, a faithful implementation of the [W3C ARIA Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).
+
+Conforming to the W3C ARIA Tabs pattern, the markup for `TabPanels` uses a container with a `tablist` role, along with styled tabs that have a `tab` role, connected to panels that have a `tabpanel` role. `TabPanels` are not intended to be used for navigation, and as such, are missing critical accessibility information for navigational behavior.
+
+`TabPanels` is a compound component. `TabPanels` is used to setup the container. `TabPanels.Tabs` is used to list the tab items. `TabPanels.Panels` is used to list the panel contents for each tab. `TabPanel.Tabs` and `TabPanels.Panels` accepts a list of `<Item />`s as their children with `key`s that must correspond to each other's tab/panel combination.
+
+### API
+
+```ts
+type TabPanels = AriaLabelingProps & {
+  /**
+   * The children of the <TabPanels> element. Should include <TabPanels.Tabs>
+   * and <TabPanels.Panels> elements.
+   */
+  children: ReactNode;
+
+  /**
+   * The initial selected key in the collection (uncontrolled).
+   */
+  defaultSelectedKey?: Key;
+
+  /**
+   * The keys of the tabs that are disabled. These tabs cannot be selected,
+   * focused, or otherwise interacted with.
+   */
+  disabledKeys?: Iterable<Key>;
+
+  /**
+   * Whether tabs are activated automatically on focus or manually.
+   *
+   * @default "automatic"
+   */
+  keyboardActivation?: "automatic" | "manual";
+
+  /**
+   * Handler that is called when the selection changes.
+   */
+  onSelectionChange?: (key: Key) => void;
+
+  /**
+   * The currently selected key in the collection (controlled).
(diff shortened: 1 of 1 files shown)


documentation/specs/TabPanels.md
```diff
@@ -0,0 +1,163 @@
+# `TabPanels` Component Specification
```
stephenjwatkins, line 1: the name, along with the previous `TabNav`, follows the Github pattern of `TabNav` and `TabPanels`.

i don't love that the consumer has to think about what version of the component to use but the implementations vary so wildly that i don't see a realistic path toward making them one (and likely why most component libraries have them separately [except for MUI (kind of)]. maybe it's good to give thought to the right pattern to use per instance, though, given that navigational behavior is critical in an app.

documentation/specs/TabPanels.md
```diff
@@ -0,0 +1,163 @@
...
+
+function BasicTabPanels() {
+  return (
+    <TabPanels
+      aria-label="History of Ancient Rome"
+      // managing selection (uncontrolled)
+      defaultSelectedKey={"one"}
+      // managing selection (controlled)
+      selectedKey={"one"}
+      onSelectionChange={(keys) => {}}
```
ranebo, line 98: I assume this is meant to be `key`

documentation/specs/TabPanels.md
```diff
@@ -0,0 +1,163 @@
+# `TabPanels` Component Specification
+
+## Overview
+
+A `TabPanels` is a set of styled links that let users switch between multiple views within a related context under the same URL.
```
ranebo, line 5: This says "a set of styled links", but that seems confusing with the below line of "should not be links that navigate the user to a new URL". Curious if they are still actually links (query params?) or something else?

stephenjwatkins: agreed that's confusing. updated

documentation/specs/TabPanels.md
stephenjwatkins: yes good catch

documentation/specs/TabPanels.md
ldewald: I agree, don't love it for needing to think about which to use but maybe we could add a note to each on usage and link to the other

documentation/specs/TabPanels.md
```diff
@@ -0,0 +1,163 @@
+# `TabPanels` Component Specification
+
+## Overview
+
+A `TabPanels` is a set of tabs that let users switch between multiple views within a related context under the same URL.
```
ldewald, line 5: **Nit:** The beginning of this sentence is slightly awkward because the component ends in `s`. I would either kill the `A` or add `component` but truly a nit comment.

documentation/specs/TabPanels.md
```diff
@@ -0,0 +1,163 @@
...
+
+---
+
+## Design
+
+`TabPanels` uses React Aria's [`useTabsList`](https://react-spectrum.adobe.com/react-aria/useTabList.html) under the hood. `useTabsList` provides the behavior and accessibility for a tab list, a faithful implementation of the [W3C ARIA Tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).
+
+Conforming to the W3C ARIA Tabs pattern, the markup for `TabPanels` uses a container with a `tablist` role, along with styled tabs that have a `tab` role, connected to panels that have a `tabpanel` role. `TabPanels` are not intended to be used for navigation, and as such, are missing critical accessibility information for navigational behavior.
+
+`TabPanels` is a compound component. `TabPanels` is used to setup the container. `TabPanels.Tabs` is used to list the tab items. `TabPanels.Panels` is used to list the panel contents for each tab. `TabPanel.Tabs` and `TabPanels.Panels` accepts a list of `<Item />`s as their children with `key`s that must correspond to each other's tab/panel combination.
```
ldewald, line 25: **Typo:** missing s
```suggestion
`TabPanels` is a compound component. `TabPanels` is used to setup the container. `TabPanels.Tabs` is used to list the tab items. `TabPanels.Panels` is used to list the panel contents for each tab. `TabPanels.Tabs` and `TabPanels.Panels` accepts a list of `<Item />`s as their children with `key`s that must correspond to each other's tab/panel combination.
```

documentation/specs/TabPanels.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+---
+
+## Behavior
+
+### Accessibility
+
+- A tab should never contain an interactive element, as doing so makes it difficult for users of assistive technologies to focus or click on the tabs. Interactive elements may live within the panels.
```
ldewald, line 154: **Thought:** Curious if there is an eslint plugin that could enforce this or maybe we could write one

documentation/specs/TabPanels.md
```diff
@@ -0,0 +1,163 @@
...
+- When the user is focused on a tab in a `TabPanels`, the following keyboard interactions apply: `Right` and `Left` arrow keys move the user between tabs. If the `keyboardActivation` is set to `automatic`, the currently focused tab's panel will open. If `keyboardActivation` is set to `manual`, the user must press `Enter` or `Space` to activate the tab.
```
ldewald, line 155: **Question:** Was this requested functionality? It's a neat idea but thinking it could create inconsistencies across the site if we're using it in different ways.

ldewald (approved):

documentation/specs/TabPanels.md
stephenjwatkins: good thought. i'll look into that

documentation/specs/TabPanels.md
stephenjwatkins: it wasn't requested functionality but it's part of the W3C ARIA Tabs pattern spec, which is why it's implemented in React Aria as such. i'm not even sure there's a way to disable it, but i understand what you mean about potential behavior inconsistency though. i'll give it some thought

PR #635, opened 2023-09-19
Create `<Stepper />` spec

OskiTheCoder:
`<Stepper />` spec only

### DIFF
--- a/documentation/specs/Stepper.md
+++ b/documentation/specs/Stepper.md
@@ -1,0 +1,102 @@
+# `Stepper` Component Specification
+
+## Overview
+
+A `Stepper` component is used to indicate progress as the user goes through a multi-step process.
+
+### Use Cases
+
+- Guide a user through a large form where parts of the form can be seperated into logical steps.
+- Guide a user through a sequence of steps where the next step depends on data from the previous step.
+
+### Features
+
+- The `orientation` property can be used to display a veritcal or horizontal stepper.
+- The `color` property can used to display a stepper with primary or inverse colors; inverse is suitable on darker backgrounds.
+- The step sequence is non-linear, which means users can navigate to previously completed steps.
+- On small screens, the step label is not visible.
+- A step that is not completed or active is considered disabled.
+
+---
+
+## Design
+
+In contrast to most Easy UI componenets, the `Stepper` component does not have an immediate `react-aria` primitive to build on top of.
+
+The `Stepper` componenet has two considerations: The outer container and individual steps. The outer container will handle orientation and spacing concerns with each individual step with respect to another. Individual steps will be represented as `Stepper.Step` and each `Stepper.Step` renders a `Stepper.StepButton` and behind the scenes a `Stepper.StepSeparator`. The `Stepper.StepButton` will make use of the `UnstyledButton` component and the clicklable area will not only include the label, but a left aligned icon as well. Representing each step as a button is necessary to support a non-linear step sequence where previously completed steps can be clicked and returned to. The `Stepper.StepSeparator` will not be exposed to the consumer, and will be omitted for the last step in the collection.
+
+### API
+
+```ts
+type StepperColor = "primary" | "inverse";
+type StepperOrientation = "horizontal" | "vertical";
+
+type Stepper = {
+  /**
+   * Color scheme that applies to stepper in various states.
+   */
+  color?: StepperColor;
+  /**
+   * Orientation of stepper, vertical replaces arrow icon with line.
+   */
+  orientation?: StepperOrientation;
+  /**
+   * The current step, other steps are either in a disabled or completed state.
+   */
+  activeStep: number;
+  /**
+   * The steps to render, individual Step components.
+   */
+  children: ReactNode;
+};
+
+type Step = {
+  /**
+   * Whether or not the step should have completed styles.
+   */
+  isCompleted: boolean;
+  /**
+   * Renders StepButton and StepSeparator.
+   */
+  children: ReactNode;
+};
+
+type StepButton = AriaButtonProps & {
+  /**
+   * StepButton content, the actual step text itself.
+   */
+  children: ReactNode;
+};
+```
+
+### Example Usage
+
+_Basic usage_:
+
+```tsx
+import { useState } from "react";
+import { Stepper } from "@easypost/easy-ui/Stepper";
+
+const steps = ["Step 1", "Step 2", "Step 3"];
+
+function App() {
+  const [activeStep, setActiveStep] = useState(0);
+  const [completed, setCompleted] = React.useState<{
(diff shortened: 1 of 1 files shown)


documentation/specs/Stepper.md
```diff
@@ -0,0 +1,102 @@
...
+- A step that is not completed or active is considered disabled.
+
+---
+
+## Design
+
+In contrast to most Easy UI components, the `Stepper` component does not have an immediate `react-aria` primitive to build on top of.
+
+The `Stepper` component has two considerations: The outer container and individual steps. The outer container will handle orientation and spacing concerns with each individual step with respect to another. Individual steps will be represented as `Stepper.Step` and each `Stepper.Step` renders a `Stepper.StepButton` and behind the scenes a `Stepper.StepSeparator`. The `Stepper.StepButton` will make use of the `UnstyledButton` component and the clickable area will not only include the label, but a left aligned icon as well. Representing each step as a button is necessary to support a non-linear step sequence where previously completed steps can be clicked and returned to. The `Stepper.StepSeparator` will not be exposed to the consumer, and will be omitted for the last step in the collection.
+
```
OskiTheCoder, line 27: `Stepper.Step` and `Stepper.StepButton` will read data from `Stepper` via context. still unsure about some of the details so i omitted that here to avoid confusion; will add later

documentation/specs/Stepper.md
```diff
@@ -0,0 +1,102 @@
...
+        <Stepper.Step key={stepLabel} isCompleted={status[index]}>
+          <Stepper.StepButton onPress={handleStep(index)}>
+            {stepLabel}
+          </Stepper.StepButton>
+        </Stepper.Step>
+      ))}
+    </Stepper>
+  );
+}
+```
```
OskiTheCoder, line 102: i wouldn't say i love this design and it is certainly subject to change during the implementation phase. other approaches considered was to have a `<Stepper />` component that accepts a "data" object similar to what exists on the site currently; in other words omitting `Stepper.Step` and `Stepper.StepButton` and everything is handled behind the scenes. maybe thats a bit easier for consumers to reason about too.

documentation/specs/Stepper.md
```diff
@@ -0,0 +1,102 @@
...
+- Guide a user through a large form where parts of the form can be separated into logical steps.
+- Guide a user through a sequence of steps where the next step depends on data from the previous step.
+
+### Features
+
+- The `orientation` property can be used to display a vertical or horizontal stepper.
+- The `color` property can be used to display a stepper with primary or inverse colors; inverse is suitable on darker backgrounds.
+- The step sequence is non-linear, which means users can navigate to previously completed steps.
+- On small screens, the step label is not visible.
+- A step that is not completed or active is considered disabled.
```
ranebo, line 18: Does disabled mean they can't click it? Or purely for styling leaving the onPress handler to determine if that step should become active?
 
What about the concept of isAccessible? If there is Step A, B, and C. And a user gets to step C, doesn't complete it, but goes back to step A, it seems like they wouldn't be able to / doesn't look like they can click step C to go back to it in order to finish it.

If not wanting to have it explicitly handled, it seems like the step after the last "completed" step should be clickable/not disabled even though it's not completed or active.

OskiTheCoder: i think my original wording was quite confusing, i updated it and your comment made me realize i think it'd be good if the component was explicit about its status options. i added a prop `status` which replaces `isComplete` and `status` can be active, accessible, disabled, or completed. depending on how implementation goes, i may do away with the `activeStep` prop since it looks like `status` should be able to capture the meaningful info on its own

documentation/specs/Stepper.md
stephenjwatkins: i like the thought put into this!

the design makes sense coming from Chakra and MUI. i think for our use case, imo, we can tighten up the API a little, since we don't have to accommodate the same kind of flexibility that the general purpose Chakra/MUI libraries have to accommodate. they're basically giving consumers building blocks that they can use as their needs require.

for us, i think we can go simpler and just have a `Stepper` and `Stepper.Step` components, basically locking down the styling/flexibility to only suit our needs. i like having those two components instead of a `data` prop personally, as i think it keeps to the spirit of the rest of our component library and provides a nice parallel in the React tree with the visuals on the screen.

inside the `Stepper.Step` component we can have any of the implementation details, like whether to render a button or separator, etc.

documentation/specs/Stepper.md
```diff
@@ -0,0 +1,104 @@
...
+   * The current step, other steps are either in a disabled or completed state.
+   */
+  activeStep: number;
+  /**
+   * The steps to render, individual Step components.
+   */
+  children: ReactNode;
+};
+
+type StepStatus = "active" | "accessible" | "completed" | "disabled";
```
stephenjwatkins, line 53: i'm slightly concerned that this state tree is a little dicey with all the different permutations that can exist lol.

for instance, does this mean two steps can be `active` at the same time? seems like the intention with the `activeStep` on the parent is to enforce only a single active step at a time. does `accessible` mean that it's clickable? would that mean a `completed` step isn't accessible/reachable?

thinking out loud...
do you think it's too heavy handed to only have an `activeStep` on the parent, and all steps previous to that are completed, and all steps after are incomplete, like the Figma below? and could we use the presence of an `onPress` handler to determine if a step is clickable?

image

documentation/specs/Stepper.md
```diff
@@ -0,0 +1,104 @@
...
+
+type Stepper = {
+  /**
+   * Color scheme that applies to stepper in various states.
+   */
+  color?: StepperColor;
+  /**
+   * Orientation of stepper, vertical replaces arrow icon with line.
+   */
+  orientation?: StepperOrientation;
```
stephenjwatkins, line 42: i assume defaults for these are `primary` and `horizontal`?

documentation/specs/Stepper.md
OskiTheCoder: good point. i think id like to see how the implementation shakes out, but i think it should be feasible to just have `Stepper` and `Stepper.Step`

documentation/specs/Stepper.md
OskiTheCoder: thinking about it, that makes sense and i believe its much more sensible to reason about. updated the API a bit to reflect this idea

documentation/specs/Stepper.md
OskiTheCoder: ah yes, updated with defaults

stephenjwatkins (approved):
i think this direction can work. as you said, i think some things will need to shake out during implementation (for instance do we need to mark some step `button`s as `disabled`, and if so, when? and i think it would be nice if `Stepper.Step` and `Stepper.StepButton` could be consolidated) but overall it seems like the right path for the API

PR #660, opened 2023-09-26
Create `<TabPanels />` component

stephenjwatkins:
## 📝 Changes

- Adds `<TabPanels />` component (the implementation of `Tabs` for navigation within the same page)
- Moves the bulk of `<TabNav />` to a common `<Tabs />` component for sharing across components

## ✅ Checklist

- [x] Changeset is added


easy-ui-react/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,94 @@
...
+  children: ReactNode;
+};
+
+/**
+ * Do not use this component directly.
+ *
+ * `Tabs` is the underlying implementation to the presentation of tabs in
+ * `TabNav` and `TabPanels`. The specifics of each implementation is handled
+ * by the respective component.
+ */
```
OskiTheCoder, line 32: **question**: not super important, but do you know if theres a doc tag or something we might be able to use here to indicate that it shouldn't be used directly?

easy-ui-react/src/TabPanels/TabPanelsTabs.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const ref = React.useRef(null);
+  const mergedProps = mergeProps(tabProps, props);
+  const state = useTabListState(mergedProps);
+  const { tabListProps } = useTabList(mergedProps, state, ref);
+
+  useEffect(
+    () => {
+      setTabListState(state);
+    },
+    // state is too comprehensive of a dependency in and of itself. only change
```
OskiTheCoder, line 38: **question**: by comprehensive, do you mean there are too many dependencies of state to include here or that the other ones aren't really useful for us?

easy-ui-react/src/Tabs/useIndicatorSizing.ts
```diff
@@ -1,30 +1,30 @@
 import { useResizeObserver } from "@react-aria/utils";
 import { MutableRefObject, useCallback, useEffect } from "react";
-import { TabNavContextType } from "./context";
+import type { TabsContextType } from "./context";
```
OskiTheCoder, line 3: **question**: is explicit type import intentional?

easy-ui-react/src/Tabs/Tabs.tsx
stephenjwatkins: good call. i updated to use `@private` and `@ignore` jsdoc tags to signify it's private (slight abuse of that tag but it's fine) and `ignore` for not being a part of public documentation

easy-ui-react/src/TabPanels/TabPanelsTabs.tsx
stephenjwatkins: yeah listening to the entire state causes a render loop because it is concerned with fields that don't apply to the tab list on this component. this just listens for the ones we care about. i updated the comment to reflect that so it's more intuitive

easy-ui-react/src/Tabs/useIndicatorSizing.ts
stephenjwatkins: it was intentional, to guard against a circular dependency import before i extracted out this hook into its own file. it's not necessary anymore though so i removed it

easy-ui-react/src/Tabs/Tabs.module.scss
```diff
@@ -68,12 +68,14 @@
...
-  transform: translateX(#{component-token("tab-nav", "indicator-position")});
-  width: component-token("tab-nav", "indicator-width");
+  transform: translateX(#{component-token("tab-nav", "indicator-position")})
+    scaleX(#{component-token("tab-nav", "indicator-width")});
+  transform-origin: left;
 
-  transition-property: transform, width;
-  transition-duration: 0.2s;
+  transition-property: transform;
+  transition-duration: 0.25s;
```
OskiTheCoder, line 79: **question** how did you decide on this as a duration? is it something from spectrum?

stephenjwatkins: i didn't see it in spectrum. it was just based on what felt right and similar to what i've seen in other libraries in the past but i'm not married to it

stephenjwatkins:
> is the height shift here just due to the height of the content changing? i presume yes but just wanted to check just in case

yeah it's from the content wrapping to two lines on the first tab

OskiTheCoder (approved):
LGTM!

PR #661, opened 2023-09-27
create <Stepper /> component

OskiTheCoder:
## 📝 Changes

Creates `<Stepper />` component

## ✅ Checklist

- [x] Changeset is added


kevinalexliu (commented):
I noticed the stepper has a bit of layout shift when accessing different steps. Is this the expected result?

beside this, code looks great!

OskiTheCoder:
> I noticed the stepper has a bit of layout shift when accessing different steps. Is this the expected result?
> 
> beside this, code looks great!

ah nope, not expected, addressed by adding a max-height on the container.

kevinalexliu (approved):
lgtm!

stephenjwatkins:
did the layout shift get resolved? in the main storybook i still see some horizontal shift, at least on my browser (chrome), when navigating between active and inactive steps:

https://github.com/EasyPost/easy-ui/assets/752942/8f0a2914-696a-4416-938a-50c29cd99e14

stephenjwatkins:
also it looks like there aren't focus states on the navigable buttons. was this intentional? it's kind of difficult when tabbing through the page on keyboard

OskiTheCoder:
> did the layout shift get resolved? in the main storybook i still see some horizontal shift, at least on my browser (chrome), when navigating between active and inactive steps:
> 
>  Screen.Recording.2023-10-02.at.11.27.47.AM.mov

i think the issue is coming from the status indicator, there is a margin that shouldn't be there on the incomplete steps, ill put up a pr to address it shortly and it should reduce the shifting. i dont think we can get rid of it fully though since the active states apply "subtitle1" font styles while the rest are "body1". you can see below what happens when i make everything "body1" and include the margin fix:

https://github.com/EasyPost/easy-ui/assets/44451175/6334ae99-3461-4e75-b82c-69d446e28590

effectively the layout shift goes away, but i dont think we can achieve this given the constraint that the active state needs to have a different font style applied.

OskiTheCoder:
> also it looks like there aren't focus states on the navigable buttons. was this intentional? it's kind of difficult when tabbing through the page on keyboard

yes, this was intentional. i inquired about the UX differences between what exists on the site currently vs what we intend to implement here and there shouldn't be any differences; this also included the "not-allowed" state for the cursor when the button is disabled

stephenjwatkins:
> yes, this was intentional. i inquired about the UX differences between what exists on the site currently vs what we intend to implement here and there shouldn't be any differences

i see. i think we should push back on that advice because it breaks accessibility in a pretty big way. was hoping to fix a lot of issues in our current site with what's here in Easy UI

OskiTheCoder:
> > yes, this was intentional. i inquired about the UX differences between what exists on the site currently vs what we intend to implement here and there shouldn't be any differences
> 
> i see. i think we should push back on that advice because it breaks accessibility in a pretty big way. was hoping to fix a lot of issues in our current site with what's here in Easy UI

im on board with that, ill put the current pr in draft and ping design about getting some focus states added

PR #666, opened 2023-10-02
fix margin and add padding around indicator

OskiTheCoder:
## 📝 Changes

- removed margin from incomplete indicator
- uniform font style

### DIFF
--- a/easy-ui-react/src/Stepper/StepButton.module.scss
+++ b/easy-ui-react/src/Stepper/StepButton.module.scss
@@ -74,7 +74,6 @@
 }
 
 .statusComplete {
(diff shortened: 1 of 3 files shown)


easy-ui-react/src/Stepper/StepButton.module.scss
```diff
@@ -130,7 +129,6 @@
 
 .indicatorIncomplete {
   border: 1px solid component-token("step-button", "incomplete-border");
-  margin: calc(#{design-token("space.0.5")} - 1px) design-token("space.0");
```
OskiTheCoder, line 133: originally added this to offset the height difference from the border, but i dont think this was working as i intended

PR #680, opened 2023-10-04
Create `<SearchNav />` spec

OskiTheCoder:
## 📝 Changes

- creates spec for `<SearchNav />` component


stephenjwatkins (approved):
looks good to me overall! i think the only thing that feels a little off is the name `LogoOptions`. not sure what a better name is though. maybe even `LogoGroup` would be slightly better to be more in parallel with `CTAGroup` below? idk, just a thought. feel free to take or leave

PR #703, opened 2023-10-16
Create `<FocusedProductLayout />` spec

stephenjwatkins:
## 📝 Changes

- Adds spec for `FocusedProductLayout`

### DIFF
--- a/documentation/specs/FocusedProductLayout.md
+++ b/documentation/specs/FocusedProductLayout.md
@@ -1,0 +1,219 @@
+# `FocusedProductLayout` Component Specification
+
+## Overview
+
+`FocusedProductLayout` defines the main content areas for a product page in a focused state.
+
+### Prior Art
+
+- [Primer `<PageLayout />`](https://primer.style/design/components/page-layout/react)
+
+---
+
+## Design
+
+`FocusedProductLayout` will be a compound component consisting of `FocusedProductLayout`, `FocusedProductLayout.Header`, and either `FocusedProductLayout.Content` or `FocusedProductLayout.WizardContent`.
+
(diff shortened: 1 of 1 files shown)


documentation/specs/FocusedProductLayout.md
```diff
@@ -0,0 +1,215 @@
...
+        renderBackArrow={(props) => {
+          // pass props that represent any unique design constraints
+          // in this case, they are `BackArrowIcon` as `children`
+          return (
+            <Link href="/account/settings" passHref>
+              <a>{props.children}</a>
+            </Link>
+          );
+        }}
+        renderLogo={() => <EasyPostLogo />}
```
stephenjwatkins, line 89: i went back and forth with these props as render props.

image

for `renderLogo`, since our library is open source, i figured it would be best to not embed our logo into the component. though the sizing constraints for the logo will be built into the logo container in this component.

for `renderBackArrow`, while most of the time this will be a client-side navigation using `Link` or another framework-specific navigation component, it could foreseeably be a button that simply changes application state, so i figured leaving flexibility to the consumer would be good. alternatively we could have specific props to enforce it being a link and/or other specific behavior, such as `backArrowHref` and `backArrowComponent` but that feels a little clunk/rigid to me.

documentation/specs/FocusedProductLayout.md
```diff
@@ -0,0 +1,215 @@
+# `FocusedProductLayout` Component Specification
```
stephenjwatkins, line 1: not entirely sure yet if this will end up being a variant of `ProductLayout` as defined previously, or a separate component like it's defined here. the API surface are has some shared properties but it also has some significant differences. i'm going to talk to design about if there are any plans for transitioning between layouts, and i think that could affect the decision path here. i anticipate fleshing this out in implementation

ralexmatthews (approved):

PR #741, opened 2023-10-27
Create `<FocusedProductLayout />`

stephenjwatkins:
## 📝 Changes

- Adds `<FocusedProductLayout />`
- Bumps `<Stepper />` breakpoint from `sm` to `md`

## ✅ Checklist

- [x] Changeset is added


easy-ui-react/src/FocusedProductLayout/FocusedProductLayout.tsx
```diff
@@ -0,0 +1,157 @@
...
+   * ```
+   *
+   * @example
+   * ```tsx
+   * <FocusedProductLayout
+   *   renderBackArrow={(props) => <button onClick={() => {}} {...props} />}
+   * />
+   * ```
+   */
+  renderBackArrow: (props: { children: ReactNode }) => ReactNode;
```
OskiTheCoder, line 50: **question** out of curiosity,  is the functionality left up to the consumer to allow for the possibility of using `next/link` and all the benefits we can get from that or is there another reason?

easy-ui-react/src/FocusedProductLayout/FocusedProductLayout.module.scss
```diff
@@ -0,0 +1,56 @@
...
+}
+
+.innerContainer {
+  display: flex;
+  flex-direction: column;
+  min-height: 100svh;
+  overflow: hidden;
+
+  @include ProductLayout.breakpoint-large-screen {
+    min-height: calc(100svh - #{design-token("space.2")});
```
OskiTheCoder, line 25: **question** not saying this should be changed or anything, more or a question out of curiosity, but does `svh` guard against anything or give us a win over setting the viewport height in a different way?

easy-ui-react/src/FocusedProductLayout/FocusedProductLayout.tsx
stephenjwatkins: yep that's precisely it. allowing the consumer to choose what happens when a user clicks "back". i left a short comment on the spec with my thoughts on the back arrow and logo render props: https://github.com/EasyPost/easy-ui/pull/703#discussion_r1360931283

let me know if you feel it should be different

easy-ui-react/src/FocusedProductLayout/FocusedProductLayout.module.scss
stephenjwatkins: good question. you probably already know about `svh`, `lvh`, and `dvh` for setting heights based on different mobile viewport configurations: https://web.dev/articles/viewport-units.

i was wanting to make sure that the layout is always at least as tall as the smallest visible mobile viewport, without stretching into the area of the browser below that (under the status bar). if it stretches into that area it causes the page to have a scrollbar. `lvh` would cause it to stretch into that bottom area, and `dvh` causes layout shift issues due to it changing the element size as the visible viewport changes.

does that make sense and/or answer the question?

OskiTheCoder (approved):
lgtm! nice work! btw, do you off the top of your head know how we can integrate this into our dashboard?

stephenjwatkins:
> do you off the top of your head know how we can integrate this into our dashboard?

that's a good question. each page ultimately controls its own layout, so i suspect we can take it page by page in the dashboard replacing the layouts accordingly. the layouts are only presentational, so they should be able to plug into any business concerns in the Next app. the Create Label page was intended to use this focused page layout, so that might be a good place to start for this particular layout

PR #750, opened 2023-10-31
feat: support base styles for UnstyledButton

OskiTheCoder:
## 📝 Changes

- support base styles for` UnstyledButton`
- update styles for Easy UI components that use `UnstyledButton`

### DIFF
--- a/easy-ui-react/src/UnstyledButton/UnstyledButton.tsx
+++ b/easy-ui-react/src/UnstyledButton/UnstyledButton.tsx
@@ -4,6 +4,8 @@ import { AriaButtonProps, mergeProps, useButton } from "react-aria";
 import { classNames } from "../utilities/css";
 import { omitReactAriaSpecificProps } from "../Button/utilities";
 
+import styles from "./UnstyledButton.module.scss";
+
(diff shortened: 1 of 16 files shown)


easy-ui-react/src/UnstyledButton/UnstyledButton.tsx
```diff
@@ -52,7 +54,7 @@ export const UnstyledButton = forwardRef<null, UnstyledButtonProps>(
         {...mergeProps(omitReactAriaSpecificProps(restProps), elementProps)}
         disabled={isDisabled}
         ref={mergeRefs(ref, inRef)}
-        className={classNames(className)}
+        className={classNames(styles.UnstyledButton, className)}
```
OskiTheCoder, line 57: i considered maybe adding it as a prop to control this behavior, but i think it makes a bit more sense as to always have this. if theres a use case im not thinking of though, im open to having a prop

stephenjwatkins: having it as a default makes sense to me as well

stephenjwatkins (approved):
nice. a lot of good cleanup

PR #762, opened 2023-11-07
docs: create `<VerticalNav />` spec

stephenjwatkins:
## 📝 Changes

- Adds spec for `<VerticalNav />` component

image

image

### DIFF
--- a/documentation/specs/VerticalNav.md
+++ b/documentation/specs/VerticalNav.md
@@ -41,25 +41,26 @@ type BaseVerticalNavProps = AriaLabelingProps & {
   };
 };
 
-export type VerticalNavProps = BaseVerticalNavProps & {
-  selectedKey: Key;
+type VerticalNavProps = BaseVerticalNavProps & {
+  selectedKey?: Key;
 };
 
-export type ExpandableVerticalNavProps = BaseVerticalNavProps & {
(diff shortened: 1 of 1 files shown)


documentation/specs/VerticalNav.md
```diff
@@ -0,0 +1,242 @@
...
+        </VerticalNav.Subnav>
+      </VerticalNav.Item>
+      <VerticalNav.Item key="3" href="/item-3" icon={Symbol} label="Item 3" />
+      <VerticalNav.Item key="4" href="/item-4" icon={Symbol} label="Item 4" />
+    </VerticalNav>
+  );
+}
+```
+
+_Expandable vertical nav:_
```
OskiTheCoder, line 111: **question**: on mobile will the popover be implemented with a menu or listbox?

OskiTheCoder (approved):
exciting stuff! lgtm

documentation/specs/VerticalNav.md
stephenjwatkins: good question. this may change after getting into the thick of implementation, but i think we can just keep it as the same implementation type (`nav`) on mobile since the behavior doesn't change from desktop to mobile. it would just be rendered as such in the container that changes to a "floating card" appearance

PR #763, opened 2023-11-07
feat: add new color alias tokens

stephenjwatkins:
## 📝 Changes

- Adds new (simplified) color alias tokens
- Updates to new tokens in Button component for an initial sampling

===

Simplify our Easy UI color token configuration.

Right now we have an overwhelming alias token setup, which carries a lot of the right patterns in general, but it hasn't scaled well due to various dynamics. We will simplify to a more basic single-layer alias abstraction over our primitive colors.

Instead of tokens such as `color.background.primary`, `color.text.subdued`, and `color.border.danger`, we'll use more simply a pattern of `color.{shade}.{weight}` like `color.primary.800`, `color.secondary.100`, and `color.neutral.500`. This allows us to keep a semantic layer, not tying our components to specific colors, while keeping the configuration more straightforward.

These will all still be reflected in our `theme/base.json` file to support dark/light mode in the future.

Note that this only deals with color tokens.


kevinalexliu (approved):
LGTM!

PR #780, opened 2023-11-14
replace isLabelVisuallyHidden with aria-label

OskiTheCoder:
## 📝 Changes

- replaces `isLabelVisuallyHidden` with `aria-label`.
- `label` is now optional.
- if `label` and `aria-label` are both omitted, a warning is logged to the console.

### DIFF
--- a/.changeset/empty-oranges-laugh.md
+++ b/.changeset/empty-oranges-laugh.md
@@ -1,0 +1,5 @@
+---
+"@easypost/easy-ui": minor
+---
+
+feat: support traditional aria labeling for TextField, Textarea, and Select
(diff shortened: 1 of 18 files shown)


stephenjwatkins (approved):
i think technically this would be a breaking (major) change but since we're in alpha mode it's fine as feat.

do you know if this is used anywhere in our front-end ecosystem?

OskiTheCoder:
> i think technically this would be a breaking (major) change but since we're in alpha mode it's fine as feat.
> 
> do you know if this is used anywhere in our front-end ecosystem?

ah yeah good point.

and yeah, `isLabelVisuallyHidden` is used in a few places. my plan is to cut a release and make the corresponding changes on the app side.

PR #828, opened 2023-12-05
fix(DataGrid): track action columns correctly

stephenjwatkins:
## 📝 Changes

- fix DataGrid using row count instead of column count to track where columns were in the table
- Adds story and test

Bug is causing gaps in tables with 1 row:

image

## ✅ Checklist

- [x] Changeset is added


ranebo (approved):
👌

PR #830, opened 2023-12-05
fix: clean up test noise

stephenjwatkins:
## 📝 Changes

- cleans up test noise
- uses correct alignment prop on Stepper
- silences console errors on tests that throw

clean test output:

image

## ✅ Checklist

- [x] Changeset is added

### DIFF
--- a/easy-ui-react/src/Stepper/Stepper.tsx
+++ b/easy-ui-react/src/Stepper/Stepper.tsx
@@ -30,11 +30,11 @@ export type StepperProps = {
 };
 
 /**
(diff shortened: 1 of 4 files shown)


easy-ui-react/src/Stepper/Stepper.tsx
```diff
@@ -30,11 +30,11 @@ export type StepperProps = {
 };
 
 /**
-* The `<Stepper />` component is used to indicate progress as the user goes 
+* The `<Stepper />` component is used to indicate progress as the user goes
```
stephenjwatkins, line 33: must have been a prettier update between then and now

PR #1309, opened 2024-08-05
Calendar component

kevinalexliu:
## 📝 Changes

Create `<Calendar />` component

## ✅ Checklist

- [x] Changeset is added

### DIFF
--- a/easy-ui-react/src/Calendar/Calendar.mdx
+++ b/easy-ui-react/src/Calendar/Calendar.mdx
@@ -1,0 +1,56 @@
+import React from "react";
+import { ArgTypes, Canvas, Meta, Controls } from "@storybook/blocks";
+import { Calendar } from "./Calendar";
+import * as CalendarStories from "./Calendar.stories";
+
+<Meta of={CalendarStories} />
+
+# Calendar
+
+Calendar display a grid of days and allow users to select a single date.
+
+<Canvas of={CalendarStories.Default}/>
+
+## Default Value
+
+A `<Calendar />` component allows users to set default selected date (uncontrolled).
+
+<Canvas of={CalendarStories.DefaultValue}/>
+
+## Set minimum and maximun value
+
+Set minimum and maximum allowed date that a user may select. Dates outside the minimum and maximum value is disabled.
+
+<Canvas of={CalendarStories.LimitAvailableDates}/>
+
+## Set Date Availability
+
+Callback that is called for each date of the calendar. If it returns true, then the date is unavailable.
+
+<Canvas of={CalendarStories.DatesAvailability}/>
+
+## Disabled Calendar
+
+Use `isDisabled` to disabled the calendar entirely.
+
+<Canvas of={CalendarStories.DisabledCalendar} />
+
+<Controls of={CalendarStories.DisabledCalendar} include={["isDisabled"]} />
+
+## Read-Only Calendar
+
+Use `isReadOnly` to make the calendar immutable.
+
+<Canvas of={CalendarStories.ReadonlyCalendar} />
+
+<Controls of={CalendarStories.ReadonlyCalendar} include={["isReadOnly"]} />
+
+## Controlled Selection
+
+For controlled selection, use `onChange` and `value`. Uncontrolled selection is possible with `defaultValue`. See associated code snippet.
+
+<Canvas of={CalendarStories.ControlledSelection} />
+
+## Properties
+
+<ArgTypes of={Calendar} />
--- a/easy-ui-react/src/Calendar/Calendar.mdx
+++ b/easy-ui-react/src/Calendar/Calendar.mdx
@@ -7,7 +7,7 @@ import * as CalendarStories from "./Calendar.stories";
 
 # Calendar
 
-Calendar display a grid of days and allow users to select a single date.
+Calendar displays a grid of days and allows users to select a single date.
 
 <Canvas of={CalendarStories.Default} />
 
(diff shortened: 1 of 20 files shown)


kevinalexliu:
will revisit in the future to add range calendar capability

kevinalexliu:
[Figma](https://www.figma.com/design/ozD467J4694z5RG5hEvQBS/Shipping-Dashboard?node-id=2688-2120&t=U1A7OF4duUVX3Fax-0)

easy-ui-react/src/Calendar/Calendar.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+## Controlled Selection
+
+For controlled selection, use `onChange` and `value`. Uncontrolled selection is possible with `defaultValue`. See associated code snippet.
+
+<Canvas of={CalendarStories.ControlledSelection} />
+
+## Invalid Selection
+
+An error message display when the selected value is invalid according to application logic.
```
stephenjwatkins, line 70: **note**: can we include text that lets the user know how to get an error message? i.e. "Select a weekend date to see an invalid selection."

easy-ui-react/src/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,94 @@
...
+  isInvalid?: boolean;
+  /**
+   * An error message to display when the selected value is invalid.
+   */
+  errorMessage?: ReactNode;
+  /**
+   * Display the days falling into the other months.
+   * @default false
+   */
+  showOutsideDays?: boolean;
```
stephenjwatkins, line 60: **suggestion**: this property name was non-obvious to understand. could we name it `showDaysOutsideCurrentMonth` (similar to MUI's https://mui.com/x/react-date-pickers/date-calendar/#show-additional-days)?

easy-ui-react/src/Calendar/Calendar.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+<Canvas of={CalendarStories.Default} />
+
+## Default Value
+
+A `<Calendar />` component allows users to set default selected date (uncontrolled).
+
+<Canvas of={CalendarStories.DefaultValue} />
+
+## Set minimum and maximun value
```
stephenjwatkins, line 20:
```suggestion
## Set minimum and maximum value
```

stephenjwatkins (changes requested):
nice work.

a few comments on the code and then a few general UX questions.

a) we need a focus indicator on the calendar cells so the user can navigate with their keyboard. keyboard navigation works but you can't tell what cell is focused.

b) on the "set minimum and maximum value" story, if you click on a date that's prior to the minimum value, it automatically selects the minimum. i personally found this a little jarring so i wondered if this was intentional/desired?

c) for "showing days outside of current month" if i click on a day outside the current month it selects that day. i assume this is intentional but the lack of hover indicator compared to the cells on the current month, makes the experience a little jarring. is there a way we can add a hover state to any clickable cell to make this more intuitive?

d) on the "set date availability" story, i can click the dates that are strikethrough and it navigates the month. this was also a bit unexpected

e) for the "readonly calendar" i would expect it to not have a hover state since the days can't be selected

f) can we double check the padding on the header vertically above/below the text? it looks a little different than what's in Figma:

image

image

easy-ui-react/src/Calendar/CalendarGrid.tsx
```diff
@@ -0,0 +1,59 @@
...
+          ))}
+        </tr>
+      </thead>
+      <tbody>
+        {[...new Array(weeksInMonth).keys()].map((weekIndex) => (
+          <tr key={weekIndex}>
+            {state
+              .getDatesInWeek(weekIndex)
+              .map((date: CalendarDate | null, i: number) =>
+                date &&
```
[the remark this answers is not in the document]
kevinalexliu: I tested it, and it seems to always have 7 dates in a week. Right now, `<td />` renders when the date is outside of the current month, and `showDaysOutsideCurrentMonth` returns `false`

easy-ui-react/src/Calendar/CalendarCell.module.scss
```diff
@@ -0,0 +1,57 @@
...
+  justify-content: center;
+  align-items: center;
+  width: component-token("calendar-cell", "space");
+  height: component-token("calendar-cell", "space");
+  cursor: pointer;
+  border-radius: design-token("shape.border_radius.lg");
+  span {
+    color: design-token("color.primary.800");
+  }
+  :focus-visible {
```
stephenjwatkins, line 19: **question**: does this need to be `&:focus-visible` to target itself, or is the idea to target all descendants with `:focus-visible`?

easy-ui-react/src/Calendar/CalendarCell.tsx
```diff
@@ -0,0 +1,64 @@
...
+      state.focusNextPage();
+    }
+    if (isPreviousMonth && !isUnavailable) {
+      state.focusPreviousPage();
+    }
+    if (!state.isInvalid(date)) {
+      state.setValue(date);
+    }
+  };
+  return (
```
stephenjwatkins, line 45:
```suggestion
  };
  
  return (
```

easy-ui-react/src/Calendar/CalendarGrid.tsx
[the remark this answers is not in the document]
stephenjwatkins: i guess i'm wondering about the `date &&` line. from my initial understanding of it, it seems like it's not doing anything since `date` will always be truthy. but it's possible i'm not understanding every case

stephenjwatkins (commented):
looks great to me. left a few minor comments.

can you send Design a link to the Storybook and make sure they say it looks good?

easy-ui-react/src/Calendar/CalendarGrid.tsx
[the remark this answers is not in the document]
kevinalexliu: It seems like few calendar systems don't have 7 days in a week (Javanese calendar), and checking `date` is to capture that.

easy-ui-react/src/Calendar/CalendarCell.module.scss
kevinalexliu: it should be `&:focus-visible`. thanks for catching this

stephenjwatkins (approved):

PR #1404, opened 2024-10-16
spec: NexusLayout

stephenjwatkins:
## 📝 Changes

- Adds spec for `NexusLayout`

image

image

Also includes primitives for nested navigation.


documentation/specs/NexusLayout.md
```diff
@@ -0,0 +1,465 @@
...
+  /** Slot for badge to be rendered. */
+  renderBadge?: () => ReactNode;
+
+  /** Slot for rendering the menu overlay. */
+  children: ReactNode;
+};
+
+// Multi-page components
+
+type NexusLayoutMultipageContainerProps = {
```
stephenjwatkins, line 88: not sure i love the name `Multipage` for the nested navigation pages but it what was called in Figma. open to suggestions.

OskiTheCoder: i cant think of a better name off the top of my head. i was thinking maybe something like `NexusLayoutSectionalContainer` but i think then we'd end up with stuff like `...SectionalSidebar` which sounds a bit odd lol. i think its fine as is

OskiTheCoder:
**question**:  do we have mockups for condensed views?

kevinalexliu (approved):

stephenjwatkins:
> **question**: do we have mockups for condensed views?

we don't have mockups for condensed views yet. in theory the API should stay the same and condensed views should be an implementation detail but it's possible it could warrant API updates. but with the pace of the project i think we'll have to evolve this layout as we get new design requirements

PR #1405, opened 2024-10-16
spec: Pill component

OskiTheCoder:
## 📝 Changes

- adds spec for supporting a `Pill` component

Screenshot 2024-10-16 at 2 47 31 PM

### DIFF
--- a/documentation/specs/Pill.md
+++ b/documentation/specs/Pill.md
@@ -62,7 +62,7 @@ _With image:_
 ```tsx
 import { Pill } from "@easypost/easy-ui/Pill";
 
-function Component() {
+function Component({ carrier }) {
   const CarrierImage = () => <Image src={`/${carrier}.png`} />;
   return <Pill icon={CarrierImage}>First Last #12345</Pill>;
 }
--- a/documentation/specs/Pill.md
+++ b/documentation/specs/Pill.md
@@ -1,0 +1,156 @@
+# `Pill` Component Specification
+
+## Overview
+
+A `Pill` is a compact element that displays contextual text typically representing selections or options; they can be dismissed as well as accompanied by an icon or image.
+
+### Use Cases
+
+- Display selections or options made by the user or system
+
+### Features
+
+- Supports icons and images
+- Supports dismissal
+
+### Risks and Challenges
+
+- Ensuring ease of use
+- Handling invalid prop configurations
+
+### Prior Art
+
+Not a 1 to 1, but similar:
+
+- [MUI `<Chip />`](https://mui.com/material-ui/react-chip/)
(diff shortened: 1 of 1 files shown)


documentation/specs/Pill.md
```diff
@@ -0,0 +1,156 @@
...
+### API
+
+```ts
+export type PillProps = {
+  /** Text label */
+  children: ReactNode;
+  /** Left aligned icon */
+  icon?: IconSymbol;
+  /** Left aligned image */
+  image?: ReactNode;
```
OskiTheCoder, line 44: will likely need to confirm the image part with design. seems like we need it per some mockups, but its not part of the figma specs for the component

ralexmatthews (approved):
Sick. Looks good to me 👍

documentation/specs/Pill.md
```diff
@@ -0,0 +1,156 @@
...
+  /** Callback function when dismissing Pill */
+  onDismiss?: () => void;
+  /** Identifier for Pill */
+  key?: Key;
```
stephenjwatkins, line 48: **nit**: since `key` is a special prop in React i think we can omit this as a prop that's unique to this component definition

documentation/specs/Pill.md
```diff
@@ -0,0 +1,156 @@
...
+function Component() {
+  return <Pill icon={SettingsIcon}>First Last #12345</Pill>;
+}
+```
+
+_With image:_
+
+```tsx
+import { Pill } from "@easypost/easy-ui/Pill";
+import SettingsIcon from "@easypost/easy-ui-icons/Settings";
```
stephenjwatkins, line 69: **suggestion**: can omit this line in this example

documentation/specs/Pill.md
stephenjwatkins, line 44: **thought**: up to you on what you think is best, but the `Icon` component i believe is designed to accept images as symbols which would allow us to use the constraints (size, color, etc) built into the `Icon` component with images, and allow the component to use a single prop.

here's the gist of how it could work:

```
type PillProps = {
  iconSymbol?: IconSymbol;
};

function Component({ carrier }) {
  // This would have to be an element passed in as a symbol
  const CarrierImage = () => <Image src={`/${carrier}.png`} />;
  return (
    <Pill iconSymbol={CarrierImage}>
      First Last #12345
    </Pill>
  );
}

<span className={className}>
  {iconSymbol && <Icon size="xs" symbol={iconSymbol} />}
  <Text>{children}</Text>
  <UnstyledButton>
    <Icon size="xs" symbol={CloseIcon} />
  </UnstyledButton>
</span>
```

stephenjwatkins (approved):

stephenjwatkins (approved):

PR #1413, opened 2024-10-17
feat: create PillGroup component

OskiTheCoder:
## 📝 Changes

- Creates `<PillGroup />` component
- Updates `border_radius` token values:
  - `sm: 3px` -> `sm: 2px`
  - `md: 5px` -> `md: 4px`

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added


easy-ui-tokens/src/shape/base.json
```diff
@@ -1,7 +1,8 @@
 {
   "shape": {
     "border_radius": {
-      "sm": { "value": "3px", "attributes": { "unit": "rem" } },
+      "xs": { "value": "3px", "attributes": { "unit": "rem" } },
```
OskiTheCoder, line 4: i guess this is technically a breaking change since there is at least one place where this is being consumed outside this project that would need to be updated. that appears to be the `Search` component from the docs site which uses `sm`. 

im open to other ways to handle this (assuming the `4px` from the Figma file is intentional)

easy-ui-react/src/Pill/Pill.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * @example
+```tsx
+* import { Pill } from "@easypost/easy-ui/Pill";
+*
+* function Component({ carrier }) {
+*   const CarrierImage = () => <Image src={`/${carrier}.png`} />;
+*   return <Pill label="First Last #12345" icon={CarrierImage} />;
+* }
+*/
+export function Pill(props: PillProps) {
```
OskiTheCoder, line 35: fwiw, i noticed `react-aria` does have a primitive for handling pills/tags with grouping https://react-spectrum.adobe.com/react-aria/useTagGroup.html. for our use case im not entirely sure if its necessary given we're not really handling explicit selection (as far as i know the only _interaction_ we're concerned with is the dismiss button), and there does seem to be a bit of overhead even with just `useTag` as it assumes you are passing along some state. our use case seems to be a bit closer to what `polaris` has https://polaris.shopify.com/components/selection-and-input/tag?example=tag-removable. 

that said, im not opposed to making this a slightly more heavy component with group/list handling because there does seem to be some wins to be had. also we could probably make it a bit flexible by allowing consumers to pass container properties similar to what we do for `HorizontalStack`, `VeritcalStack`..etc, but im also iffy on the way it recommends to handle list state as just feels a bit heavy. idk just thinking out loud lol

easy-ui-tokens/src/shape/base.json
stephenjwatkins: good question. our border radiuses today feel a little arbitrary. it might be good to ask Rachel if there are a set of border radiuses that she uses in the designs/would like to systematize, and we can hopefully adjust once and for all

easy-ui-react/src/Pill/Pill.tsx
stephenjwatkins: this is a good point actually. it probably would be good to use React Aria since it does provide some accessibility patterns that would be hard to build ourselves (like keyboard nav). we should maybe consider using `TagGroup` as the name since this should typically be used in a group context

it may be an opportunity to use React Aria Components to make it easier: https://react-spectrum.adobe.com/react-aria/TagGroup.html. with this, i think the work would be mostly limited to styling

easy-ui-tokens/src/shape/base.json
OskiTheCoder: updated `border_radius` mappings per discussion with rachel. should be documented now in Figma.

easy-ui-react/src/Pill/Pill.tsx
OskiTheCoder: updated component architecture and used `react-aria-components`

easy-ui-react/src/PillGroup/PillGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+};
+
+function Pill(props: PillProps) {
+  const { label, icon } = props;
+
+  return (
+    <Tag textValue={label} className={styles.Pill} {...props}>
+      {(renderProps: TagRenderProps) => {
+        return (
+          <>
```
stephenjwatkins, line 137: **nit**: you could could return the parenthesis content directly without the manual return statement

easy-ui-react/src/PillGroup/PillGroup.tsx
```diff
@@ -0,0 +1,155 @@
...
+    ...getResponsiveDesignToken("pill-group", "gap", "space", gap),
+    ...getComponentToken("pill-group", "align", align),
+    ...getComponentToken("pill-group", "block-align", blockAlign),
+    ...getComponentToken("pill-group", "wrap", wrap ? "wrap" : "nowrap"),
+    ...getComponentToken(
+      "pill-group",
+      "display",
+      inline ? "inline-flex" : "flex",
+    ),
+  } as React.CSSProperties;
```
stephenjwatkins, line 110: **thought**: i feel like we'll want some kind of helper to work with our primitives more easily, especially with react aria components. i'll think about that some more but open to ideas

stephenjwatkins (approved):
looks good. thanks for converting to TagGroup. i think it's much better

can you check that the bundle sizes for files are not much higher with/without react aria components? i think on build it shows the output

OskiTheCoder:
> can you check that the bundle sizes for files are not much higher with/without react aria components? i think on build it shows the output

checked and the bundle sizes look not much higher with react-aria-components. there was this comment logged however,

build_comments

issue was raised to the adobe folks here https://github.com/adobe/react-spectrum/issues/6933

OskiTheCoder:
> looks good. thanks for converting to TagGroup. i think it's much better

agree and react aria components was pretty pleasant to work with overall, definitely wouldn't be against leveraging it wherever it makes sense, even for existing components tbh

stephenjwatkins (approved):
> checked and the bundle sizes look not much higher with react-aria-components. there was this comment logged however,
> 
> issue was raised to the adobe folks here [adobe/react-spectrum#6933](https://github.com/adobe/react-spectrum/issues/6933)

got ya. looks like it'll be fixed when they upgrade their Parcel version

PR #1422, opened 2024-10-22
T79394 Create `<SplitToggleCard />` component

kevinalexliu:
## 📝 Changes

- Create `<SplitToggleCard />` component

## ✅ Checklist

- [ ] ~Design tokens are utilized~
- [x] Specs are up-to-date
- [x] Changeset is added

### DIFF
--- a/easy-ui-react/src/SplitToggleCard/SplitToggleCard.stories.tsx
+++ b/easy-ui-react/src/SplitToggleCard/SplitToggleCard.stories.tsx
@@ -7,7 +7,7 @@ import { VerticalStack } from "../VerticalStack";
 type Story = StoryObj<typeof SplitToggleCard>;
 
 const meta: Meta<typeof SplitToggleCard> = {
-  title: "Components/SplitToggleCard",
+  title: "Components/Cards/SplitToggleCard",
   component: SplitToggleCard,
   args: {
     isDisabled: false,
@@ -27,12 +27,11 @@ const Template = (args: SplitToggleCardProps) => {
   return (
     <SplitToggleCard {...args} aria-labelledby={id}>
       <VerticalStack gap="1">
-        <Text variant="subtitle2" color="primary.800">
+        <Text id={id} variant="subtitle2" color="primary.800">
           Toggle Label
         </Text>
-        <div>
+        <Text variant="caption">
           <Text
-            variant="caption"
             color={
               args.isSelected || args.defaultSelected
                 ? "positive.600"
@@ -41,11 +40,11 @@ const Template = (args: SplitToggleCardProps) => {
           >
             {status}
           </Text>
-          <Text variant="caption" color="primary.800">
+          <Text color="primary.800">
             : AI-based Delivery Timelines for each Shipping Option at Checkout
             for Shoppers.
           </Text>
-        </div>
+        </Text>
       </VerticalStack>
     </SplitToggleCard>
   );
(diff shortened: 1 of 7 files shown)


easy-ui-react/src/SplitToggleCard/SplitToggleCard.stories.tsx
```diff
@@ -0,0 +1,107 @@
+import { Meta, StoryObj } from "@storybook/react";
+import React, { useState, useId } from "react";
+import { SplitToggleCard, SplitToggleCardProps } from "./SplitToggleCard";
+import { Text } from "../Text";
+import { VerticalStack } from "../VerticalStack";
+
+type Story = StoryObj<typeof SplitToggleCard>;
+
+const meta: Meta<typeof SplitToggleCard> = {
+  title: "Components/SplitToggleCard",
```
stephenjwatkins, line 10: **question**: can we put this and `Card` under `Components/Cards/`?

easy-ui-react/src/SplitToggleCard/SplitToggleCard.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const Template = (args: SplitToggleCardProps) => {
+  const id = useId();
+  const status =
+    args.isSelected || args.defaultSelected ? "Enabled" : "Disabled";
+  return (
+    <SplitToggleCard {...args} aria-labelledby={id}>
+      <VerticalStack gap="1">
+        <Text variant="subtitle2" color="primary.800">
```
stephenjwatkins, line 30: **issue**: this should have the id attached, `id={id}`, so that the `aria-labelledby` is connected to this text

easy-ui-react/src/SplitToggleCard/SplitToggleCard.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const [selected, setSelected] = useState(false);
+    const status = selected ? "Enabled" : "Disabled";
+    return (
+      <SplitToggleCard
+        isSelected={selected}
+        onChange={setSelected}
+        aria-labelledby={id}
+      >
+        <VerticalStack gap="1">
+          <Text variant="subtitle2" color="primary.800">
```
stephenjwatkins, line 68: **issue**: this should have the id attached, `id={id}`, so that the `aria-labelledby` is connected to this text

easy-ui-react/src/SplitToggleCard/SplitToggleCard.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+            Toggle Label
+          </Text>
+          <div>
```
stephenjwatkins, line 71: **suggestion**: i would suggest having this `<div>` (and the other story) be `<Text variant="caption">` and then the `Text` components underneath it just setting the `color`. `Text` is built to be wrapped as much as needed

as it is right now the line height is off because it's using the parent's line height

image

easy-ui-react/src/SplitToggleCard/SplitToggleCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * ```
+ */
+
+export const SplitToggleCard = (props: SplitToggleCardProps) => {
+  const { children, ...toggleProps } = props;
+
+  return (
+    <Card.Container>
+      <HorizontalGrid columns={["80px", "1fr"]}>
+        <Card.Area background="secondary" padding="1.5">
```
stephenjwatkins, line 68: **issue**: i think this would be better represented as `24px` padding:

image

stephenjwatkins (commented):
nice work. a few comments

easy-ui-react/src/SplitToggleCard/SplitToggleCard.stories.tsx
kevinalexliu: good call, addressed

stephenjwatkins (approved):

PR #1427, opened 2024-10-23
Spinner spec

kevinalexliu:
## 📝 Changes

Add Spinner spec

### DIFF
--- a/documentation/specs/Spinner.md
+++ b/documentation/specs/Spinner.md
@@ -1,0 +1,118 @@
+# `Spinner` Component Specification
+
+## Overview
+
+A `Spinner` component indicate the loading state of a component or page.
+
+### Use Cases
+
+- Use a `Spinner` component to display a visible loading indicator for situations when an asynchronous API call or process might take a while.
+
+### Features
+
+- Supports `md` and `lg` sizes.
+- Supports optional label.
+- Supports various colors.
+- It's animated.
+
+### Prior Art
+
+- [Paste `<Spinner />`](https://paste.twilio.design/components/spinner)
+
+---
+
+## Design
+
+### API
+
+```ts
+type SpinnerProps = {
+  /**
+   * Size of spinner.
+   * @default md
+   */
+  size?: "md" | "lg";
+  /**
+   * The label for spinner.
+   */
+  children?: string;
+  /**
+   * Adjust color of spinner and label.
+   * @default "neutral.500"
+   */
+  color?: ThemeTokenNamespace<"color.text">;
+};
+```
+
+### Example Usage
+
+```tsx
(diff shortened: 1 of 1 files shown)


documentation/specs/Spinner.md
```diff
@@ -0,0 +1,118 @@
...
+type SpinnerProps = {
+  /**
+   * Size of spinner.
+   * @default md
+   */
+  size?: "md" | "lg";
+  /**
+   * The label for spinner.
+   */
+  children?: string;
```
stephenjwatkins, line 38: **suggestion**: i think `children` should be a `ReactNode` so that it will allow for `Text` within

documentation/specs/Spinner.md
```diff
@@ -0,0 +1,118 @@
...
+      aria-label={children ? undefined : "loading"}
+      aria-labelledby={children ? undefined : id}
+    >
+      <div className={styles.spinnerCircle} />
+      {children && (
+        <Text id={id} color={color}>
+          {children}
+        </Text>
+      )}
+    </div>
```
stephenjwatkins, line 101: **nit**: up to you, you could simplify this a little by using `visuallyHidden`, something like:

```tsx
<div
  className={styles.spinner}
  style={style}
  role="alert"
  aria-labelledby={id}
>
  <div className={styles.spinnerCircle} />
    <Text id={id} color={color} visuallyHidden={!children}>
      {children ? children : "Loading"}
    </Text>
</div>
```

documentation/specs/Spinner.md
```diff
@@ -0,0 +1,118 @@
...
+}
+```
+
+---
+
+## Behavior
+
+### Accessibility
+
+- Spinner always has the `role="alert"` attribute.
```
stephenjwatkins, line 112: **suggestion**: i would consider using `role="status"`

image

stephenjwatkins (approved):
approved but with a couple thoughts

documentation/specs/Spinner.md
OskiTheCoder: +1

documentation/specs/Spinner.md
```diff
@@ -0,0 +1,118 @@
+# `Spinner` Component Specification
```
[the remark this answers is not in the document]
stephenjwatkins: today it is filled to a fixed percentage and then rotates, but good point in that in the future it might be determinate. maybe it'd be worth using this:

https://react-spectrum.adobe.com/react-aria/useProgressBar.html

i notice they give a circle example and the spectrum progress circle uses that as an underlying primitive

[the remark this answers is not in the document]
kevinalexliu: I'll add this during implementation

OskiTheCoder (approved):

PR #1430, opened 2024-10-24
feat: create ToggleCard component

OskiTheCoder:
## 📝 Changes

- Creates `<ToggleCard />` component

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added

### DIFF
--- a/easy-ui-react/src/ToggleCard/ToggleCard.tsx
+++ b/easy-ui-react/src/ToggleCard/ToggleCard.tsx
@@ -148,6 +148,8 @@ function ToggleCardBody(props: ToggleCardBodyProps) {
   );
 }
 
+/** Represents the header content in a `<ToggleCard />`*/
 ToggleCard.Header = ToggleCardHeader;
 
+/** Represents the body content in a `<ToggleCard />`*/
 ToggleCard.Body = ToggleCardBody;
--- a/easy-ui-react/src/ToggleCard/ToggleCard.tsx
+++ b/easy-ui-react/src/ToggleCard/ToggleCard.tsx
@@ -1,8 +1,10 @@
-import React, { ReactNode } from "react";
+import React, { ReactNode, useMemo } from "react";
 import { Card } from "../Card";
 import { HorizontalStack } from "../HorizontalStack";
-import type { ToggleProps } from "../Toggle";
-import { Toggle } from "../Toggle";
+import {
+  flattenChildren,
+  getDisplayNameFromReactNode,
+} from "../utilities/react";
 
 import styles from "./ToggleCard.module.scss";
 
@@ -22,9 +24,8 @@ export type ToggleCardProps = {
  * It is a simple compound component consisting of `<ToggleCard.Header />`
  * and `<ToggleCard.Body />`.
  *
- * `<ToggleCard.Header />` automatically handles the rendering for the `Toggle`
- * control and accepts the corresponding props. Consumers only need to pass in
- * the content to render alongside the `Toggle` control.
+ * `<ToggleCard.Header />` should handle the rendering for the `Toggle`
+ * control and the content to render alongside it.
  *
  * `<ToggleCard.Body />` is responsible for rendering the body content.
  *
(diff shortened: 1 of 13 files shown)


stephenjwatkins (commented):
it all looks good but i wonder if we should adjust the API to where the `ToggleCard.Header` can just accept a `Toggle` and `Text` directly. that way it simplifies the API, leaves it a little more flexible, and the `toggleChildren` feels a little awkward to use. the position of the toggle then is up to the consumer as to how they position it: `<ToggleCard.Header><Toggle /><Text /></ToggleCard.Header>`. this component would manage the overall card styling/constraints but it would just be a lightweight wrapper

OskiTheCoder:
> it all looks good but i wonder if we should adjust the API to where the `ToggleCard.Header` can just accept a `Toggle` and `Text` directly. that way it simplifies the API, leaves it a little more flexible, and the `toggleChildren` feels a little awkward to use. the position of the toggle then is up to the consumer as to how they position it: `<ToggleCard.Header><Toggle /><Text /></ToggleCard.Header>`. this component would manage the overall card styling/constraints but it would just be a lightweight wrapper

thats good with me, will make those adjustments

easy-ui-react/src/ToggleCard/ToggleCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+    for (let i = 0; i < headerChildren.length; i++) {
+      if (getDisplayNameFromReactNode(headerChildren[i]) === "Toggle") {
+        foundToggle = true;
+        break;
+      }
+    }
+    if (!foundToggle) {
+      throw new Error("ToggleCard.Header must contain Toggle");
+    }
+  }, [children]);
```
OskiTheCoder, line 105: can remove if this feels a bit odd. i also thought about changing up the name of this component (VerticalActionCard?) if it doesn't necessarily have to render a toggle, but also im fine with it as is too given that there arent plans to support anything other than a toggle yet

stephenjwatkins: i like the error handling. yeah we can evolve the name if we find it has more uses later

easy-ui-react/src/ToggleCard/ToggleCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  return (
+    <Card.Area background="neutral.050" padding="0.5">
```
stephenjwatkins, line 108: looking at the design i think we need to add `paddingX` and `paddingY` props that are the same signature as `padding` but allow for different left/right and top/bottom padding values on the `Card`. would it make sense to include in this task or follow up later?

easy-ui-react/src/ToggleCard/ToggleCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+ *    <Toggle aria-labelledby="some id" isReadOnly />
+ *  </ToggleCard.Header>
+ *  <ToggleCard.Body>body</ToggleCard.Body>
+ * </ToggleCard>
+ * ```
+ */
+export function ToggleCard(props: ToggleCardProps) {
+  const { children } = props;
+
+  return <Card.Container>{children}</Card.Container>;
```
stephenjwatkins, line 81: looking at the design seems we need to accept a `borderRadius` prop here with our token values similar to boxShadow. would it make sense to include in this task or follow up later?

easy-ui-react/src/ToggleCard/ToggleCard.tsx
OskiTheCoder: good point. it would remove the random `div` i needed to add in the body lol. 

i think we should do it as a follow up. ill create a task and assign it to myself

easy-ui-react/src/ToggleCard/ToggleCard.tsx
OskiTheCoder: oh interesting, didnt even catch that the border radius is totally different, ill add it to the same ticket

stephenjwatkins (approved):

PR #1487, opened 2024-11-22
feat: Menu supports multiple selection

kevinalexliu:
## 📝 Changes

- Menu multiple selection 

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added

### DIFF
--- a/easy-ui-react/src/Menu/Menu.mdx
+++ b/easy-ui-react/src/Menu/Menu.mdx
@@ -72,7 +72,7 @@ A `<Menu />` supports multiple selection mode by using `selectionMode` prop on `
 
 <Canvas of={MenuStories.MultipleSelection} />
 
-A multiple selection `<Menu />` supports Select All checkbox with `all` as key on `<Menu.Item />`.
+A multiple selection `<Menu />` supports Select All checkbox with `all` as key on `<Menu.Item />`. When a "select all" event occurs—either through clicking the Select All menu item or using `cmd` + `a` keyboard shortcut— `onSelectionChange(keys)` will be passed `all` as its key instead of a list of all keys.
 
 <Canvas of={MenuStories.MultipleSelectionWithSelectAll} />
 
--- a/easy-ui-react/src/Menu/Menu.mdx
+++ b/easy-ui-react/src/Menu/Menu.mdx
@@ -72,9 +72,7 @@ A `<Menu />` supports multiple selection mode by using `selectionMode` prop on `
 
 <Canvas of={MenuStories.MultipleSelection} />
 
-A multiple selection `<Menu />` supports Select All checkbox using `all` as key on `<Menu.Item />`.
-
-**Important**: Select all checkbox only works with `all` as its key, and when all options are selected, `selectedKeys` returns `all` instead of every single key.
+A multiple selection `<Menu />` supports Select All checkbox with `all` as key on `<Menu.Item />`.
 
 <Canvas of={MenuStories.MultipleSelectionWithSelectAll} />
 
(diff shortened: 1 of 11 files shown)


easy-ui-react/src/Menu/utilities.ts
```diff
@@ -36,3 +40,55 @@ export function getUnmergedPopoverStyles(
...
+      }
+    }
+  }, [
+    selectedKeys,
+    isFocused,
+    isSelectAll,
+    selectedSize,
+    itemSize,
+    state.selectionManager,
+  ]);
```
stephenjwatkins, line 91: **suggestion**: i'm concerned with having the state be modified through a connection with the rendering cycle via `useEffect` and using the menu item's `isFocused`. that paradigm often opens up a component to race conditions and fragility in the future since it's harder to guard against related changes that could affect the behavior.

i would suggest instead for this to be purely event based. this could be done by extending the `onAction` function in `MenuOverlay` and running the selection logic in there based on the SELECT_ALL key. let me know if that makes sense. i can help walk you through the idea.

stephenjwatkins (commented):
i notice that the cursor changes from `pointer` to `default` when hovering between the menu item and the label around the checkbox. we should make sure the cursor is the same. either forcing the label to be `pointer` in the menu or menu items forcing to be `default`

easy-ui-react/src/Menu/MenuOverlay.tsx
```diff
@@ -105,8 +134,44 @@ function MenuOverlayContent<T extends object>(props: MenuOverlayProps<T>) {
...
+          // Select Select ALL checkbox when all other checkboxes are selected
+          selectionManager.selectedKeys.size ===
+            getItemSize(menuTreeState) - 2 &&
+          !selectionManager.isSelected(key)
+        ) {
+          selectionManager.selectAll();
+        }
+        break;
+    }
+  };
```
stephenjwatkins, line 162: we'll still want to call `onAction(key)` in this handler

easy-ui-react/src/Menu/Menu.stories.tsx
```diff
@@ -230,3 +231,58 @@ export const CustomPlacement: StoryObj<MenuOverlayProps<unknown>> = {
...
+
+    return (
+      <Menu>
+        <Menu.Trigger>
+          <DropdownButton>Click me</DropdownButton>
+        </Menu.Trigger>
+        <Menu.Overlay
+          onSelectionChange={setSelectedKeys}
+          selectionMode="multiple"
+          selectedKeys={selectedKeys}
```
stephenjwatkins, line 249: could you add an action handler to these multiselection stories so we can see the selection result in the actions panel? i.e. `onAction={action("Selected")}`

stephenjwatkins:
@kevinalexliu posted an example of what works in a React Aria compatible way: https://github.com/EasyPost/easy-ui/pull/1507

check it out and let me know what you think. if it works, feel free to approve and merge. it'll go into this branch. also has some other minor Menu fixes

easy-ui-react/src/Menu/Menu.mdx
```diff
@@ -66,6 +66,16 @@ A `<Menu />` can be placed on different edges of its trigger element. See the co
 
 <Controls of={MenuStories.CustomPlacement} include={["placement"]} />
 
+## Multiple Selection
+
+A `<Menu />` supports multiple selection mode by using `selectionMode` prop on `<Menu.Overlay />`. `onSelectionChange` and `selectedKeys` (controlled) are used to make selection.
+
+<Canvas of={MenuStories.MultipleSelection} />
+
+A multiple selection `<Menu />` supports Select All checkbox with `all` as key on `<Menu.Item />`.
```
stephenjwatkins, line 75:
```suggestion
A multiple selection `<Menu />` supports Select All checkbox with `all` as key on `<Menu.Item />`. When a "select all" event occurs—either through clicking the Select All menu item or using `cmd` + `a` keyboard shortcut— `onSelectionChange(keys)` will be passed `all` as its key instead of a list of all keys.
```

PR #1542, opened 2024-12-18
ModeSwitcher and BreadcrumbNavigation draft

OskiTheCoder:


easy-ui-react/src/ModeSwitcher/ModeSwitcher.module.scss
```diff
@@ -0,0 +1,29 @@
...
+  align-items: center;
+  justify-content: space-between;
+  gap: design-token("space.1");
+  padding: design-token("space.2");
+  border: design-token("shape.border_width.1") solid
+    theme-token("color.neutral.300");
+  border-radius: design-token("shape.border_radius.md");
+  cursor: pointer;
+  width: 100%;
+  max-width: 160px;
```
OskiTheCoder, line 16: set this as `max-width` as a magic number instead of `133px` because i noticed with `133px` the icon resizing caused the arrow to shrink with longer text content

Uploading Screen Recording 2024-12-17 at 5.50.09 PM.mov…

easy-ui-react/src/ModeSwitcher/ModeSwitcher.module.scss
```diff
@@ -0,0 +1,29 @@
...
+  border-color: theme-token("color.neutral.800");
+}
+
+.popover {
+  border: design-token("shape.border_width.1") solid
+    theme-token("color.neutral.300");
+  border-radius: design-token("shape.border_radius.md");
+  background-color: white;
+  padding: design-token("space.2");
+}
```
OskiTheCoder, line 29: there are some open design questions i have regarding the yellow color thats given in figma and what we have available in easy-ui and also the spacing within the popover / radio items

easy-ui-react/src/ModeSwitcher/ModeSwitcher.tsx
```diff
@@ -0,0 +1,144 @@
...
+export function FakeTopLevelForgeModeSwitcher() {
+  const [mode, setMode] = useState("production");
+
+  return (
+    <FakeForgeComponent mode={mode as "production" | "test"}>
+      <ModeSwitcher onModeChange={setMode} />
+    </FakeForgeComponent>
+  );
+}
+
```
OskiTheCoder, line 27: @stephenjwatkins does this align with how we could potentially use this component? the top level Forge component accepts `mode` as a prop and the dispatch function is passed directly to `ForgeLayout.ModeSwitcher` ?

PR #1544, opened 2024-12-18
feat: DatePicker and DateRangePicker

kevinalexliu:
## 📝 Changes

- Fix calendar styles
- Add DatePicker and DateRangePicker

## ✅ Checklist

- [ ] ~Visuals match Design Specs in Figma~(There is no design spec)
- [x] Specs are up-to-date
- [x] Changeset is added


easy-ui-react/src/DatePicker/DatePickerTrigger.tsx
```diff
@@ -0,0 +1,80 @@
...
+          )}
+          <InputIcon
+            alignment="end"
+            icon={CalendarMonth}
+            size={size}
+            isDisabled={isDisabled}
+          />
+        </UnstyledButton>
+      </div>
+      {(isInvalid || state.isInvalid) && (
```
OskiTheCoder, line 73: **question**: are there instances where `isInvalid` and `state.isInvalid` would be different?

stephenjwatkins (commented):
can you ensure the sizes are the same as the `Select` and `TextField`?

looks like it's about `8px` bigger and the calendar icon isn't vertically center on my browser:

image

similarly the `sm` size calendar icon isn't vertically centered on my screen:

image

easy-ui-react/src/DatePicker/DatePickerTrigger.tsx
kevinalexliu: `state.isInvalid` is determined by `useDatePickerState`. For example, it handles end date before start date error and provides a default error message. 

Users pass `isInvalid` props to `DatePicker` based on their business logic.

easy-ui-react/src/Calendar/CalendarGrid.module.scss
```diff
@@ -1,6 +1,8 @@
 @use "../styles/common" as *;
 .CalendarGrid {
   border-collapse: collapse;
+  z-index: 999;
```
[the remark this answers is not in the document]
kevinalexliu: after looking at it again, `z-index` should be on the `diaglog` to show above the `popover` instead of Calendar.

easy-ui-react/src/DatePicker/DatePicker.module.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.dialog {
+  background: theme-token("color.neutral.000");
+  border-radius: design-token("shape.border_radius.lg");
+  box-shadow: design-token("shadow.modal");
+  pointer-events: auto;
+  outline: none;
+  position: relative;
+  z-index: design-token("z-index.input_icon");
```
OskiTheCoder, line 45: **suggestion**: up to you and a minor thing, but i think we could just use the value of 1 here or define a component level token. _generally_ so far we've been assigning z-index tokens in association with some semantics

stephenjwatkins (approved):

PR #1570, opened 2025-01-14
fix(DataGrid): align actions menu to end

stephenjwatkins:
## 📝 Changes

- In cases where the actions cell is expanded beyond its content width, align it toward the end 

before:

image

after:

image


ralexmatthews (approved):
🚀

PR #1614, opened 2025-02-24
fix(Button): remove support for className

stephenjwatkins:
## 📝 Changes

- remove support for overwriting `className` on `Button`

`<Button className="extend" />` overwrites button styles by accidentally spreading props at the end. this is a pretty big departure from Easy UI's current expectations that presentational components contain their styles and provide extensions only through props, and can cause further breaking changes later with a harder upgrade path

🚨 this is technically a breaking change but due to being in pre-release mode, it's better served as a patch. we will need to update 4 instances in our products. these can be found with this `<Button([\s\n\r]+|[^>]+)className` regex 🚨 

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added

### DIFF
--- a/.changeset/brown-buckets-whisper.md
+++ b/.changeset/brown-buckets-whisper.md
@@ -1,0 +1,5 @@
+---
+"@easypost/easy-ui": patch
+---
+
+fix(Button): extend className instead of overwriting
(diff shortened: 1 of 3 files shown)


ralexmatthews (approved):
Looks good to me 👍

OskiTheCoder (approved):
LGTM! what a sneaky one lol

PR #1704, opened 2025-04-10
HostedUILayout spec

OskiTheCoder:
## 📝 Changes

`HostedUILayout` component spec

### DIFF
--- a/documentation/specs/HostedUILayout.md
+++ b/documentation/specs/HostedUILayout.md
@@ -1,0 +1,248 @@
+# `HostedUILayout` Component Specification
+
+## Overview
+
+`HostedUILayout` defines the header, main content, and multi-page content areas of a HostedUI product page.
+
+### Prior Art
+
+- [Primer `<PageLayout />`](https://primer.style/design/components/page-layout/react)
+- [Paste `<SidebarNavigation /`>](https://paste.twilio.design/components/sidebar-navigation)
+
+---
+
+## Design
+
+`HostedUILayout` will be a compound component consisting of several pieces including `HostedUILayout`, `HostedUILayout.Header`, `HostedUILayout.Content`, and `HostedUILayout.MultipageContainer`.
+
+In structure and style, it will be very similar to the existing `HostedUILayout` with some key differences. For instance, the default presentation for `HostedUILayout` will be the multi-page navigational layout.
+
+An additional point worth highlighting is that the `MultipageContainer` will be constructed in a way where it can be exported as a stand-alone piece and usable in a non `HostedUILayout` context.
+
+### API
+
+```ts
+type HostedUILayoutProps = {
+  /** Layout children. */
+  children: ReactNode;
+};
+
+type HostedUILayoutHeaderProps = {
+  /** Header children */
+  children: ReactNode;
+};
+
+type HostedUILayoutLogoGroupProps = {
+  /** LogoGroup children */
+  children: ReactNode;
+};
+
+type HostedUILayoutContentProps = {
+  children: ReactNode;
+};
+
+type HostedUILayoutActionsProps = {
+  /** Actions children. */
+  children: ReactNode;
+};
+
+type HostedUILayoutActionBadgeProps = {
+  /** Badge children. */
+  children?: ReactNode;
+};
+
+type HostedUILayoutMenuActionProps = {
+  /** Optional custom accessibility label describing the menu action. */
+  accessibilityLabel?: string;
+  /** Icon symbol for the action. */
+  iconSymbol: IconSymbol;
+  /** Slot for badge to be rendered. */
+  renderBadge?: () => ReactNode;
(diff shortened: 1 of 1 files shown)


documentation/specs/HostedUILayout.md
```diff
@@ -0,0 +1,249 @@
...
+
+---
+
+## Design
+
+`HostedUILayout` will be a compound component consisting of several pieces including `HostedUILayout`, `HostedUILayout.Header`, `HostedUILayout.Content`, and `HostedUILayout.MultipageContainer`.
+
+In structure and style, it will be very similar to the existing `NexusLayout` with some key differences. For instance, the default presentation for `HostedUILayout` will be the multi-page navigational layout.
+
+An additional point worth highlighting is that the `MultipageContainer` will be constructed in a way where it can be exported as a stand-alone piece and usable in a non `HostedUILayout` context.
```
OskiTheCoder, line 20: this is to support the "parent" view for managing sub accounts in forge

documentation/specs/HostedUILayout.md
```diff
@@ -0,0 +1,249 @@
...
+                <Button size="sm" variant="outlined">
+                  Edit Account Settings
+                </Button>
+              </HostedUILayout.MultipageHeader>
+              <div>Content</div>
+            </HostedUILayout.MultipageContent>
+          </HostedUILayout.MultipageInnerContainer>
+        </HostedUILayout.MultipageContainer>
+      </HostedUILayout.Content>
+    </HostedUILayout>
```
OskiTheCoder, line 227: open to suggestions if any of these names could be better

ranebo: Hmm I think the name makes sense. I guess the weird thing possible is that for the Parent Hosted UI sub account manage page would be in the forge layout itself? That's what I'm assuming anyways, would that cause any issues?

ranebo (approved):

documentation/specs/HostedUILayout.md
```diff
@@ -0,0 +1,249 @@
...
+          >
+            <Menu.Overlay onAction={() => {}}>
+              <Menu.Item>Action 1</Menu.Item>
+              <Menu.Item>Action 2</Menu.Item>
+            </Menu.Overlay>
+          </HostedUILayout.MenuAction>
+        </HostedUILayout.Actions>
+      </HostedUILayout.Header>
+      <HostedUILayout.Content>
+        <HostedUILayout.MultipageContainer>
```
stephenjwatkins, line 183: given that this `Multipage` section will live in contexts outside of a hosted UI, i wonder if this should be its own component under layouts. i.e. `MultipageSection`. then it can be rendered within `HostedUILayout.Content`

documentation/specs/HostedUILayout.md
```diff
@@ -0,0 +1,249 @@
...
+          <HostedUILayout.MultipageBrandHeader>
+            <SomeOtherLogo />
+          </HostedUILayout.MultipageBrandHeader>
+          <HostedUILayout.MultipageInnerContainer>
+            <HostedUILayout.MultipageSidebar>
```
stephenjwatkins, line 188: just checking will this support instances without a `Sidebar`? i think i see some sidebar-less designs in Figma

documentation/specs/HostedUILayout.md
```diff
@@ -0,0 +1,249 @@
...
+- `HostedUILayout.MultipageSidebarNav` will be rendered as `nav`
+- `HostedUILayout.MultipageSidebarNavLink` will render as `<a>`
+- Selected nav links will be decorated as `aria-current="page"`
+
+### Dependencies
+
+- `Card`
+- `Text`
+- `useLink`
+- Will necessitate extending `EasyUIProvider` with navigation hooks to support client-side links. See [client side routing](https://react-spectrum.adobe.com/react-aria/routing.html#routerprovider).
```
stephenjwatkins, line 249: nit: i think this isn't applicable anymore since it was added previously

documentation/specs/HostedUILayout.md
OskiTheCoder: yeah the idea is it would be its own stand alone component and be able to slot into forge

`import { MultipageContainer } from "@easypost/easy-ui/MultipageContainer"`

documentation/specs/HostedUILayout.md
OskiTheCoder: thats a good idea, will opt for that approach instead

documentation/specs/HostedUILayout.md
OskiTheCoder: yep thats correct, i asked if the sidebar section is needed at all, still a pending question

documentation/specs/HostedUILayout.md
OskiTheCoder: ah good catch, will return to update this file as i begin implementation

PR #1755, opened 2025-05-08
Adds SupportAgent, ReceiptLong, Graph3, and BarChart icons

ranebo:
## 📝 Changes

Adds SupportAgent, ReceiptLong, Graph3, and BarChart icons. Removes the unused Grade Icon as it was removed from material-symbols

## ✅ Checklist

- [x] Changeset is added

### DIFF
--- a/easy-ui-icons/src/Graph3.json
+++ b/easy-ui-icons/src/Graph3.json
@@ -1,0 +1,5 @@
+{
+  "name": "graph_3",
+  "style": "outlined",
+  "source": "@material-symbols/svg-300"
+}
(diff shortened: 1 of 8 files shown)


easy-ui-icons/src/Graph3.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "grade",
+  "name": "graph_3",
```
Justintime50, line 2: thought: if the reference for this in easy-ui is also changing from graph to graph_3 (or Grade to Graph3), then this release would be a breaking change and per semver need to be major version bump. Has easyui hit a major release yet or it it still alpha/beta?

Justintime50 (approved):

easy-ui-icons/src/Graph3.json
ranebo: Yeah still "pre-release" right now, I believe we are ok with it per the slack thread

PR #1766, opened 2025-05-15
feat: add InkSelection icon

stephenjwatkins:
## 📝 Changes

- adds icon in support of feature work
- also adjusts package version in package lock


ralexmatthews (approved):
🖊️

PR #1794, opened 2025-06-10
rename HostedUILayout to CustomerPortalLayout

OskiTheCoder:
## 📝 Changes

- renames `HostedUILayout` component to `CustomerPortalLayout` to align with product name change
- will have a follow up diff in the web app to update the imports/names


easy-ui-react/src/CustomerPortalLayout/CustomerPortalLayout.module.scss
```diff
@@ -1,8 +1,8 @@
 @use "../styles/common" as *;
 @use "../styles/unstyled";
 
-.HostedUILayout {
-  @include component-token("hostedui-layout", "max-width", 1369px);
+.CustomerPortalLayout {
```
Justintime50, line 4: Everything looks good but wondered about the casing of this change, intentional to be upper cased? (Looking on phone so can't see where it's used)

Justintime50 (dismissed):

Justintime50 (approved):

PR #1887, opened 2026-05-18
Fix in CalendarCell for date range selection

ranebo:
## 📝 Changes

Fix in CalendarCell for date range selection. Date range was no longer getting selected, the check never returned the proper boolean to actually set the value. Updated to no longer have custom handling (not sure how it was working before really) and just allow the underlying react-aria to handle the selection.

## ✅ Checklist

- [X] Visuals match Design Specs in Figma
- [X] Stories accompany any component changes
- [X] Code is in accordance with our style guide
- [X] Design tokens are utilized
- [X] Unit tests accompany any component changes
- [X] TSDoc is written for any API surface area
- [X] Specs are up-to-date
- [X] Console is free from warnings
- [X] No accessibility violations are reported
- [X] Cross-browser check is performed (Chrome, Safari, Firefox)
- [X] Changeset is added

### DIFF
--- a/.changeset/open-friends-fall.md
+++ b/.changeset/open-friends-fall.md
@@ -1,0 +1,5 @@
+---
+"@easypost/easy-ui": patch
+---
+
+Fixes CalendarCell date range selection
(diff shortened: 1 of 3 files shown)


stephenjwatkins (approved):
Seems like a reasonable fix to me

PR #1912, opened 2026-08-25
feat(MultiSelect): open dropdown when input receives focus

stephenjwatkins:
Set `menuTrigger="focus"` on the underlying ComboBox so clicking or tabbing into the input opens the dropdown. Previously the options only appeared when the user typed or clicked the arrow button.


ralexmatthews (approved):
🔥

PR #1915, opened 2026-08-28
feat: honor Provider's navigate prop in link components

stephenjwatkins:
## 📝 Changes

Makes Easy UI in line with supporting client side routing via `RouterProvider`:
https://react-spectrum.adobe.com/v3/routing.html

`<Provider navigate useHref />` populates React Aria's router context, but that context is only read by hooks that call `useRouter()`/`useLinkProps()`. `useButton()` isn't one of them — it renders `href` as a plain attribute — and several nav components build their anchors by hand. Those components did a full page load even when a `navigate` function was configured.

Adds `useRouterLinkProps()`, which resolves `href` through the provider's `useHref` and returns an `onClick` that delegates to `navigate` via React Aria's `handleLinkClick`. Wires it into `<UnstyledButton />` (covering `<Button />`, `<IconButton />`, `<DropdownButton />`, `<KebabButton />`, `<PlanCard.Button />`, `<Pagination />`, `<SearchNav />` CTAs, and `<ForgeLayout.BackButton />`), `<TabNav.Item />`, `<VerticalNav.Item />`, `<VerticalNav.Subnav.Item />`, and `<VerticalNav.SupplementaryAction />`. These components also accept `routerOptions` now.

`<Menu.Item href />` already client navigated through `useMenuItem()`, but re-applied the raw `href` on top of the resolved one; it now respects `useHref` too.

Backwards compatible: without a `navigate` function React Aria's router context reports itself as native and `handleLinkClick` is a noop, so links behave exactly as before. Cross-origin links, `target="_blank"`, downloads, and modifier-clicked links keep loading normally. The `as`/`hrefComponent` escape hatches still work and take precedence over the provider's router, so apps can migrate a route at a time.

Adds `ClientSideRouting` stories for `<Button />`, `<TabNav />`, `<VerticalNav />`, and `<Menu />` that wire the components to a fake router through `<Provider />`, plus the docs sections that render them.

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added

### DIFF
--- a/.changeset/curly-lions-travel.md
+++ b/.changeset/curly-lions-travel.md
@@ -1,0 +1,30 @@
+---
+"@easypost/easy-ui": minor
+---
+
+Honor `<Provider navigate />` in components that render links
+
+`<Provider navigate useHref />` populates React Aria's router context, but most
+Easy UI components built their anchors with `useButton()` or by hand, neither of
+which reads that context. Those components did a full page load even when a
+`navigate` function was configured.
+
+`<Button />`, `<IconButton />`, `<DropdownButton />`, `<KebabButton />`,
+`<TabNav.Item />`, `<VerticalNav.Item />`, `<VerticalNav.Subnav.Item />`,
+`<VerticalNav.SupplementaryAction as="a" />`, and everything else built on
+`<UnstyledButton />` (`<PlanCard.Button />`, `<SearchNav />` CTAs,
+`<Pagination />`, `<ForgeLayout.BackButton />`) now navigate through the
+provider's `navigate` function and resolve their `href` through its `useHref`.
+`<Menu.Item href />` already navigated client-side, but rendered the unresolved
+`href`; it now respects `useHref` as well.
+
+These components also accept `routerOptions` for passing router-specific options
+through to `navigate`.
+
+This is backwards compatible. Without a `navigate` function React Aria's router
+context reports itself as native and the new click handling is a noop, so links
+behave exactly as before. Links to other origins, links with `target="_blank"`,
+download links, and modifier-clicked links continue to load normally even when a
+`navigate` function is configured. The `as`/`hrefComponent` props remain
+supported for components that need a specific link component, and take
+precedence over the provider's router.
--- a/.changeset/curly-lions-travel.md
+++ b/.changeset/curly-lions-travel.md
@@ -35,7 +35,8 @@ declare module "@react-types/shared" {
 Separately, `onClick` on `<Button />` and everything else built on
 `<UnstyledButton />` fired twice per click, because React Aria's `useButton()`
 already calls it and the original handler was spread onto the element alongside
-it. It now fires once.
+it. It now fires once, as a plain DOM handler, so `preventDefault()` reaches the
+real click on both pointer and keyboard activation.
 
 Apps that don't pass `navigate` are unaffected. React Aria's router context
 reports itself as native, the new click handling is a noop, and links behave
@@ -47,6 +48,12 @@ to `navigate`, so a link that has to reach the server—a download, a sign out, 
 path served by another app on the same origin—needs to say so. Links to other
 origins, links with `target="_blank"` or `download`, modifier-clicked links, and
 links whose `onClick` calls `preventDefault()` continue to load natively. So do
-disabled links, which render without an `href`. The `as`/`hrefComponent` props
-remain supported for components that need a specific link component, and take
-precedence over the provider's router.
+disabled links, which render without an `href`.
+
+The `as`/`hrefComponent` props remain supported for components that need a
+specific link component. On `<TabNav.Item />` and the `<VerticalNav />` items
+they take precedence over the provider's router, which is left out entirely.
+`<Menu.Item hrefComponent />` is the exception: it receives the `href` as
+written, but React Aria's menu item handles links internally and can't be opted
+out of, so the click reaches `navigate` as well. Pair it with a router rather
+than using it instead of one.
--- a/.changeset/curly-lions-travel.md
+++ b/.changeset/curly-lions-travel.md
@@ -21,6 +21,11 @@ provider's `navigate` function and resolve their `href` through its `useHref`.
 These components also accept `routerOptions` for passing router-specific options
 through to `navigate`.
 
+Separately, `onClick` on `<Button />` and everything else built on
+`<UnstyledButton />` fired twice per click, because React Aria's `useButton()`
+already calls it and the original handler was spread onto the element alongside
+it. It now fires once.
+
 Apps that don't pass `navigate` are unaffected. React Aria's router context
 reports itself as native, the new click handling is a noop, and links behave
 exactly as before.
--- a/.changeset/curly-lions-travel.md
+++ b/.changeset/curly-lions-travel.md
@@ -19,7 +19,18 @@ provider's `navigate` function and resolve their `href` through its `useHref`.
 `href`; it now respects `useHref` as well.
 
 These components also accept `routerOptions` for passing router-specific options
-through to `navigate`.
+through to `navigate`. Following React Aria, `routerOptions` is typed as `never`
+until an app declares what its router accepts:
+
+```tsx
+import type { NavigateOptions } from "react-router";
+
+declare module "@react-types/shared" {
+  interface RouterConfig {
+    routerOptions: NavigateOptions;
+  }
+}
+```
 
 Separately, `onClick` on `<Button />` and everything else built on
 `<UnstyledButton />` fired twice per click, because React Aria's `useButton()`
--- a/.changeset/curly-lions-travel.md
+++ b/.changeset/curly-lions-travel.md
@@ -35,8 +35,9 @@ declare module "@react-types/shared" {
 Separately, `onClick` on `<Button />` and everything else built on
 `<UnstyledButton />` fired twice per click, because React Aria's `useButton()`
 already calls it and the original handler was spread onto the element alongside
-it. It now fires once, as a plain DOM handler, so `preventDefault()` reaches the
-real click on both pointer and keyboard activation.
+it. It now fires exactly once per activation, pointer or keyboard. On links it
+stays a plain DOM handler, so `preventDefault()` reaches the real click the
+router reads.
 
 Apps that don't pass `navigate` are unaffected. React Aria's router context
 reports itself as native, the new click handling is a noop, and links behave
--- a/.changeset/curly-lions-travel.md
+++ b/.changeset/curly-lions-travel.md
@@ -21,10 +21,16 @@ provider's `navigate` function and resolve their `href` through its `useHref`.
 These components also accept `routerOptions` for passing router-specific options
 through to `navigate`.
 
-This is backwards compatible. Without a `navigate` function React Aria's router
-context reports itself as native and the new click handling is a noop, so links
-behave exactly as before. Links to other origins, links with `target="_blank"`,
-download links, and modifier-clicked links continue to load normally even when a
-`navigate` function is configured. The `as`/`hrefComponent` props remain
-supported for components that need a specific link component, and take
+Apps that don't pass `navigate` are unaffected. React Aria's router context
+reports itself as native, the new click handling is a noop, and links behave
+exactly as before.
+
+Apps that already pass `navigate` keep working, but the links listed above stop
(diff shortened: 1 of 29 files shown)


ralexmatthews (dismissed):
Nice work 🔥 I went through it and didn't really notice anything. I also had claude do a sweep and it found some minor things if you wanted to take a look.

<details>
<summary>Claude Review:</summary>

--- 
⏺ Code review(high · 5 findings)
easy-ui-react/src/Menu/MenuItem.tsx
  ●  98 [correctness] Omitting "href" makes Menu.Item pass the useHref-resolved href to hrefComponent, so router-aware link components double-prefix base paths.
easy-ui-react/src/Button/Button.mdx
  ●  82 [docs]        The docs example passing routerOptions is a type error for consumers because RouterOptions resolves to never without a RouterConfig module augmentation that is never documented.
easy-ui-react/src/Provider/Provider.tsx
  ● 133 [correctness] EasyUIRouterProvider only mounts React Aria's RouterProvider when navigate is truthy, so passing useHref alone is a silent no-op — now load-bearing after this PR.
.changeset/curly-lions-travel.md
  ●  21 [docs]        The changeset claims SearchNav CTAs accept routerOptions, but CTAItemProps is plain AriaButtonProps<"button">, so the prop is a type error there.
easy-ui-react/src/Button/utilities.ts
  ●  54 [correctness] onClick is absent from omitReactAriaSpecificProps, so it is spread raw onto the element and also invoked by usePress, firing the handler twice per click.

Review done — no correctness bugs in the new hook or its wiring. The core mechanism checks out: handleLinkClick gets the raw href while the DOM gets the useHref-resolved one, mergeProps doesn't let undefined clobber real values, the router onClick chains after any user onClick so preventDefault() still wins, and disabled links are excluded. Full vitest suite (504 tests), tsc --build, and eslint all clean.

5 low-severity findings, above. The two worth acting on before merge:

- MenuItem.tsx:98 — Menu.Item + hrefComponent now receives the resolved href, so a router-aware link component double-prefixes base paths. It also has no as-style router bypass, which contradicts the changeset's precedence claim.
- Button.mdx:82 — the documented routerOptions example doesn't compile without a RouterConfig module augmentation the docs never mention (the PR's own tests cast around it).

The other three are doc accuracy (SearchNav CTAs don't accept routerOptions), a silent no-op (useHref without navigate), and a pre-existing double-fire of onClick that this PR makes load-bearing by documenting it as the opt-out.
</summary>

ralexmatthews:
<details>
<summary>Claudes update:</summary>

---
⏺ Code review(high · 5 findings)
easy-ui-react/src/Button/utilities.ts
  ●  53 [correctness]   Omitting "onClick" routes the consumer's handler solely through usePress, which on keyboard activation fabricates a synthetic MouseEvent, so preventDefault() no longer suppresses client-side navigation.
easy-ui-react/src/Menu/MenuItem.tsx
  ●  83 [correctness]   Omitting href/routerOptions from the item props passed to useMenuItem has no effect, because useMenuItem re-reads the node from the collection rather than using the props it is handed.
easy-ui-react/src/Provider/Provider.test.tsx
  ● 298 [test-coverage] The test's CustomLink defines its own onClick after spreading linkProps, overriding the forwarded handler, so useMenuItem's click handler never runs and the assertion cannot detect the bug above.
easy-ui-react/src/Provider/Provider.tsx
  ● 133 [correctness]   EasyUIRouterProvider only mounts React Aria's RouterProvider when navigate is truthy, so passing useHref alone is a silent no-op — unchanged from the prior review.
easy-ui-react/src/SearchNav/CTAGroup.tsx
  ●  71 [correctness]   SecondaryCTAItem now accepts and honors routerOptions in the desktop rendering, but the condensed/overflow path rebuilds items as <Menu.Item href target> and drops it.

Head moved cd69b1a → d74ce34 (4 new commits, clean fast-forward, +139/−16). One prior finding is genuinely fixed, but two of the fixes don't do what they claim — and one of them is a regression, so the PR is worse off than before on that axis. Tests pass (506), tsc --build and eslint clean.

Prior findings: routerOptions docs — fixed (the augmentation snippet is correct and verified to unlock the prop on all five components). onClick double-fire — fixed, but see below. Menu.Item href resolution — the rendered href is now raw, but the click side isn't. SearchNav routerOptions — types fixed, condensed rendering still drops it. useHref without navigate — untouched, still open.

The two to look at before merge:

- utilities.ts:53 (regression, eb974f6) — adding "onClick" to the omit list means the handler only reaches usePress, which on keyboard activation synthesizes a fake MouseEvent. So preventDefault() no longer suppresses navigation: mouse click correctly skips navigate, but Enter and Space both call navigate("/x"). Since the changeset documents that preventDefault() opt-out as the supported escape hatch, a keyboard user now client-navigates where a mouse user doesn't. Fixing the double-call is right; it just can't cost the keyboard path.
- MenuItem.tsx:83 (753d67b) — the omit(item.props, ["href","routerOptions"]) indirection is inert. useMenuItem ignores the props it's handed and re-reads the node from the collection, so the provider's navigate still fires for hrefComponent items and the link component's own navigation never runs. Only the href omit at line 111 is doing real work. The new test at Provider.test.tsx:298 passes for the wrong reason — its CustomLink overrides the forwarded onClick instead of chaining it, so it can't catch this; chaining it the way next/link does makes the assertion fail.
</details>

ralexmatthews:
<details>
<summary>Claude followup</summary>
---

⏺ Code review(high · 5 findings)
easy-ui-react/src/UnstyledButton/UnstyledButton.tsx
  ● 56 [correctness] Passing onClick: undefined into useButton() removes the consumer's handler from the only keyboard activation path, so onClick never fires on Enter/Space for buttons without an href.
  ● 56 [consistency] A disabled link button correctly suppresses onPress and navigation but still invokes the consumer's onClick, so "disabled" is enforced for one handler prop and not the other.
easy-ui-react/src/Menu/Menu.mdx
  ● 87 [docs]        Menu.mdx claims an hrefComponent that navigates itself will navigate twice while the changeset says to pair it with a router; the measured behavior matches neither.
easy-ui-react/src/SearchNav/CTAGroup.tsx
  ● 74 [correctness] routerOptions is now forwarded when CTAs are rebuilt as Menu.Items, but the sibling RouterLinkProps — download, hrefLang, ping, referrerPolicy — are still dropped.
easy-ui-react/src/Provider/Provider.tsx
  ● 20 [correctness] Passing useHref without navigate remains a no-op; documented in TSDoc, which is a reasonable resolution.

Re-measured on the correct deps. Two of my earlier claims were stale-environment artifacts, not PR problems:

- The failing CTAGroup test — on react-aria@3.51.0 user.click opens the overflow menu and it passes. Suite is green: 63 files / 511 passed / 2 skipped. CI was right all along.
- The "15 pre-existing tsc errors" I reported every pass — tsc --build --force now gives 0 errors. Those were the old react-aria missing MappedDateValue plus RefObject variance from the stale @react-types.

One blocker survives, and it got sharper. UnstyledButton.tsx:56 — onClick on a non-link button doesn't fire on keyboard. Confirmed on 3.51.0: <Button onClick> with no href gives mouse 1, Enter 0, Space 0, virtual 1; IconButton the same. onPress fires in every cell, for a prop React Aria's types define as an alias for onPress. It's also platform-dependent — shouldPreventDefaultKeyboard() short-circuits on isMac() && key === 'Enter', so Windows/Linux keyboard users get nothing on either key while macOS users get Enter but not Space.

The A/B across revisions shows no version has satisfied both properties at once — each fix traded one for the other:

┌──────────┬──────────────────────────────────────────────┬────────────────────────────────┐
│ revision │ non-link onClick (mouse/Enter/Space/virtual) │ link preventDefault → navigate │
├──────────┼──────────────────────────────────────────────┼────────────────────────────────┤
│ main     │ 2 / 1 / 1 / 2                                │ 0 / 0 / 0 / 0                  │
├──────────┼──────────────────────────────────────────────┼────────────────────────────────┤
│ d74ce34  │ 1 / 1 / 1 / 1                                │ 0 / 1 / 1 / 0                  │
├──────────┼──────────────────────────────────────────────┼────────────────────────────────┤
│ 953d258  │ 1 / 0 / 0 / 1                                │ 0 / 0 / 0 / 0                  │
└──────────┴──────────────────────────────────────────────┴────────────────────────────────┘

Scoping the withholding to links fixes it — onClick: href ? undefined : props.onClick into useButton(), plus dropping the raw onClick from the merge in the non-link branch. That was verified, not just proposed: every cell becomes 1 on both Win32 and MacIntel, links keep navigate 0 under preventDefault, suite stays green. It was reverted; tree is clean at 953d258.

Remaining non-blocking: the Menu.mdx/changeset contradiction (behavior re-confirmed — it navigates once via the provider with the unresolved href, so neither wording is accurate), the dropped download/hrefLang/ping/referrerPolicy in rebuilt SearchNav menu items, and useHref-without-navigate as a documented no-op.

</details>

ralexmatthews:
<details>
<summary>Claude followup</summary>

---
⏺ Code review(high · 4 findings)
easy-ui-react/src/UnstyledButton/UnstyledButton.tsx
  ● 62 [correctness]   Routing onClick through usePress breaks preventDefault() for type="submit"/"reset" on keyboard activation, because react-aria leaves the native click intact there but invokes the handler with a fabricated MouseEvent.
easy-ui-react/src/UnstyledButton/UnstyledButton.test.tsx
  ● 91 [test-coverage] The new keyboard test only asserts the onClick call count, so it passes even when the handler's preventDefault() is inert against a fabricated event.
easy-ui-react/src/Menu/Menu.mdx
  ● 87 [docs]          Menu.mdx says an hrefComponent that navigates itself will navigate twice and to avoid pairing it with navigate; measurement shows it navigates once via the provider with the hrefComponent bypassed.
easy-ui-react/src/SearchNav/CTAGroup.tsx
  ● 74 [correctness]   Mirroring CTAs into Menu.Items forwards only href/target/routerOptions, dropping the other RouterLinkProps that CTAItemProps publicly accepts.

The original problem is genuinely fixed — and their fix is the one that was verified (they used lodash/omit where I used onClick: undefined; functionally identical). Every cell of the matrix is now 1 on both Win32 and MacIntel, and links keep navigate 0/0/0/0 under preventDefault. Both properties hold together for the first time in this PR. Suite green: 63 files / 515 passed / 2 skipped, tsc 0 errors, eslint clean.

But there's a fifth regression in the same fake-event family, now scoped to type="submit"/type="reset". usePress only kills the native click for type="button" — shouldPreventDefaultUp() returns false for submit/reset. So the real click still fires while the consumer's handler receives triggerSyntheticClick()'s fabricated MouseEvent, and preventDefault() on the fake event can't stop the real one:

┌──────────────────────────────┬─────────┬──────────────────────────────────────┐
│           revision           │ onClick │ onSubmit (mouse/Enter/Space/virtual) │
├──────────────────────────────┼─────────┼──────────────────────────────────────┤
│ main                         │ 2/2/2/2 │ 0/0/0/0                              │
├──────────────────────────────┼─────────┼──────────────────────────────────────┤
│ 953d258                      │ 1/1/1/1 │ 0/0/0/0                              │
├──────────────────────────────┼─────────┼──────────────────────────────────────┤
│ a010185                      │ 1/1/1/1 │ 0/1/1/0                              │
├──────────────────────────────┼─────────┼──────────────────────────────────────┤
│ plain <button type="submit"> │ 1/1/1/1 │ 0/0/0/0                              │
└──────────────────────────────┴─────────┴──────────────────────────────────────┘

So a submit button doing client-side validation (if (!valid) e.preventDefault()) submits anyway on Enter or Space but not on click. FormLayout.mdx:44 recommends exactly that pattern for keyboard submission, and FormLayout.stories.tsx:249 uses it. type="reset" is the same (onReset 0/1/1/0). Regression against both main and 953d258.

Fix is a one-line widening of the same condition, verified to leave the rest of the matrix and the suite intact:

const keepsNativeClick =
  Boolean(href) || props.type === "submit" || props.type === "reset";

Also worth flagging: the new keyboard test at UnstyledButton.test.tsx:91 asserts only the call count, so it structurally cannot catch this — the handler does run, its preventDefault() is just inert. Asserting a suppressed onSubmit would.

Still open from before: the Menu.mdx:87 wording (the changeset's version is the accurate one) and the dropped download/hrefLang/ping/referrerPolicy in rebuilt SearchNav menu items. useHref-without-navigate is now properly documented and I'd drop it as a finding.
</details>

ralexmatthews (approved):
LGTM 👍

PR #1920, opened 2026-09-02
feat(DataGrid, Pagination): sticky footer with numbered pagination and rows per page

stephenjwatkins:
## 📝 Changes

Adds a sticky footer to `<DataGrid />` and the pagination pieces it's built from.

**`<DataGrid />`**

- `renderFooter` renders a bar pinned to the bottom of the scroll container, modeled on the existing sticky header — it reuses the edge interceptors for its shadow-on-scroll and pins horizontally so it stays put while rows scroll sideways.
- `<DataGrid.Footer />` lays the bar out with purely positional `start`, `center`, and `end` slots, so it stays unaware of what it holds and the data grid isn't tied to any one pagination scheme.
- `<DataGrid.Pagination />` and `<DataGrid.RowsPerPage />` are fully controlled presets for the common case. Either can be swapped for a custom control, or `renderFooter` can skip `<DataGrid.Footer />` entirely.
- Sticky shadows are lightened (opacity `0.5` → `0.35`), which affects the frozen column shadows as well as the new footer's.

**`<Pagination />`**

- `<Pagination.Pages />` is a numbered scheme alongside the existing `<Pagination.Dropdown />`, for when the user should be able to see and reach nearby pages directly. The scheme is inferred from the child rather than configured. Optional `First`/`Last` jump buttons keep the ends of a long range reachable once it truncates.
- `<Pagination.RowsPerPage />` is a fully controlled page-size menu. It's a sibling of `<Pagination />` rather than a child, since that slot takes a scheme and the two sit side by side in design.
- Both take a `size` of `sm` or `md`, resolved to the nearest tokens on the space scale.

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added

### DIFF
--- a/.changeset/datagrid-sticky-footer.md
+++ b/.changeset/datagrid-sticky-footer.md
@@ -2,4 +2,4 @@
 "@easypost/easy-ui": minor
 ---
 
-Adds a sticky footer to `<DataGrid />` through `renderFooter`, along with `<DataGrid.Footer />`, `<DataGrid.Pagination />`, and `<DataGrid.RowsPerPage />`. Adds a numbered pagination scheme to `<Pagination />` through `<Pagination.Pages />` and a rows per page menu through `<Pagination.RowsPerPage />`, both in two sizes through `size`. Also lightens `<DataGrid />`'s sticky shadows, which affects the frozen column shadows as well as the new footer's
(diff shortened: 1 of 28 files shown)


jallen-cyber (approved):
:fire:

PR #1934, opened 2026-09-14
chore(deps): upgrade Storybook from v9 to v10

stephenjwatkins:
## 📝 Changes

Upgrades Storybook `9.1.13` → `10.6.0` ([SHPE-1675](https://simplerpostage.atlassian.net/browse/SHPE-1675)). Storybook 10 is ESM-only, which drives most of this.

- **`.storybook/main.ts` is now real ESM.** `getAbsolutePath()` used `require.resolve`, and v10 leaves the CJS globals undefined. The helper existed for Yarn PnP / monorepo resolution, but v10 only *requires* resolved paths for `core.builder` and local addons — neither of which we use — so it's deleted in favor of bare package names rather than ported to `import.meta.resolve` (which returns a URL, not a path).
- **Shared tsconfigs move off `moduleResolution: "node"`**, which v10 dropped support for along with its `typesVersions` fields. `base.json` → `"bundler"`, node variants → `"nodenext"`, with `module` set alongside each.
- **`UserEvent` now imports from the `@testing-library/user-event` root** in 9 test files. They used `dist/types/setup/setup`, a deep path the package's `exports` map doesn't expose, so it stopped resolving under `"bundler"`. The root import already matched `utilities/test.ts` and `MultiSelect.test.tsx`.
- **`eslint-config-easy-ui` no longer applies `plugin:storybook/recommended`.** The root flat config already spreads `storybook.configs`; once the two plugin copies deduped to one instance, ESLint failed outright with `Cannot redefine plugin "storybook"`. The mismatched `9.1.13`/`0.12.0` pair had been masking the duplicate registration.
- **CI moves to Node 24 via `.nvmrc`.** All three workflows pinned `20.x`, but Node 20 hit EOL on 2026-04-30 and stops getting security patches — including `release.yml`, which publishes to npm. Workflows now use `node-version-file: ".nvmrc"` so contributors and CI can't drift. `engines.node` drops the EOL `^20.19.0` branch, leaving `>=22.12.0`.
- **`style:` commit applies prettier 3.9 formatting** to 14 files. Refreshing the lockfile moved prettier `3.6.2` → `3.9.6` (already allowed by `^3.6.0`), and 3.9 collapses short unions and normalizes markdown emphasis. Formatting only, kept in its own commit so the Storybook diff stays reviewable.

Also fixes a latent bug found on the way in: `utilities/storybook.tsx` imported `Decorator` from `@storybook/react`, which isn't declared in any workspace and resolved only transitively.

I removed `@storybook/addon-mcp`, which `storybook upgrade` added unprompted — a new addon isn't part of a v9→v10 upgrade.

### Verification

`build`, `lint`, 591 unit tests, and `build:storybook` all green, on a clean `npm ci`. Dev server boots on v10.6.0 serving 446 entries with no console errors, and no ESM or deprecation warnings in either build.

Node 24 isn't installed locally, so I verified on **22.22.3 and 25.9.0** to bracket it — CI's run here is the direct confirmation.

### Reviewer notes

⚠️ **Chromatic diffs need a deliberate walk.** A major Storybook upgrade re-renders every story, so expect diffs even where no component changed. Please don't rubber-stamp the baselines — a genuine regression could ride along in the noise.

## ✅ Checklist

- [x] ~~Visuals match Design Specs in Figma~~
- [x] ~~Stories accompany any component changes~~
- [x] ~~Design tokens are utilized~~
- [x] ~~Unit tests accompany any component changes~~
- [x] ~~TSDoc is written for any API surface area~~
- [x] ~~Specs are up-to-date~~
- [ ] Cross-browser check is performed (Chrome, Safari, Firefox)
- [x] ~~Changeset is added~~ — devDependency/tooling only, no published behavior change

[SHPE-1675]: https://simplerpostage.atlassian.net/browse/SHPE-1675?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ralexmatthews (dismissed):
Niiiiiice LGTM 👍

Just one small suggestion, if prettier 3.6 -> 3.9 includes some breaking changes, should we go ahead and bump the package.json one to 3.9 as well?

ralexmatthews (approved):
🔥

PR #1939, opened 2026-09-17
feat(Popover): add a general-purpose Popover

stephenjwatkins:
## 📝 Changes

Adds a general-purpose `<Popover />`, plus a public `<SelectButton />` so a field-shaped trigger can be attached to it.

Every overlay in Easy UI was welded to a collection role — `Menu` to `menu`, `Select` to `listbox`, `DatePicker` to its own private `dialog` — so a consumer who needed arbitrary content anchored to a trigger either reached for `Modal`, which blocks the page, or hand-rolled a seventh copy of the positioning sandwich. `ForgeLayoutControls` is the cautionary case: a one-off with a magic `crossOffset={116}` and no accessible name.

### `<Popover />`

```tsx
<Popover>
  <Popover.Trigger>
    <SelectButton>Add Packaging</SelectButton>
  </Popover.Trigger>
  <Popover.Overlay width={480}>
    <Popover.Header>
      <Popover.Title>Shipment Packaging</Popover.Title>
    </Popover.Header>
    <Popover.Body>{/* anything */}</Popover.Body>
    <Popover.Footer>
      <Button>Packaging Complete</Button>
    </Popover.Footer>
  </Popover.Overlay>
</Popover>
```

- **`<Popover.Trigger />` is the only wiring path.** It clones its single child with `mergeProps`, so any `forwardRef` component that spreads rest props works — every Easy UI button, and a consumer's own component. A consumer's `onPress` on the trigger survives.
- **Content can skip the slots.** Handed to `<Popover.Overlay />` directly, it gets the padding and nothing else. With `<Popover.Header />` / `<Popover.Body />` / `<Popover.Footer />`, the header and footer stay pinned and only the body scrolls.
- **`<Popover.Title />` supplies the accessible name**, and warns in development when there's neither a title nor an `aria-label`.
- **`usePopoverTrigger()`** lets content inside the overlay close it, the same way `useModalTrigger()` does.
- **The surface is a `<Box />`**, so `<Popover.Overlay />` also accepts every `<Box />` style prop.

### `react-aria-components`, not the hooks

This is the first component to choose RAC deliberately rather than incidentally, so **ADR 014** records why. Both standing objections dissolve on inspection: nesting a RAC `<Dialog>` makes RAC's `isDialog` check deterministically false, so our dialog owns `role="dialog"`; and providing `OverlayTriggerStateContext` makes RAC defer to our state, which is what lets the house clone-child trigger stay instead of RAC's `DialogTrigger`.

In exchange, both `DismissButton`s, the underlay, `data-placement`, and `--trigger-width` come for free. That last one removes the `useTriggerWidth` + `triggerWidth === null` first-paint gate that `MenuOverlay` and `SelectOverlay` both carry, so `width="trigger"` costs no dropped first commit.

Two mechanics are load-bearing and indirect, and both have tests pinning them:

- **Focus containment** arrives through `Dialog` → `useDialog` → `useOverlayFocusContain` → the ancestor `Overlay`'s `FocusScope`. RAC's own `shouldContainFocus` is `false` here.
- **`<Popover />` inside `<Modal />`** is safe only because of `useOverlay`'s module-level topmost-overlay stack. `ModalUnderlay` would otherwise read a portaled popover click as an outside press. Same mechanism that makes `Select`-in-`Modal` work today, and the thing most likely to regress on a react-aria bump.

### Surface tokens — please sanity-check against Figma

Three existing anchored surfaces disagreed, so this was settled by counting: `Menu`, `Select`, and `MultiSelect` all render the same surface out of `Menu/_mixins.scss`, so `Popover` renders it too — white, `border_radius.md`, a 1px `neutral.300` border, `space.2` padding. A fourth look was the one outcome worth avoiding. `DatePicker`'s heavier `border_radius.lg` + `shadow.modal` treatment stays its own thing.

**One value couldn't match.** The dropdown family uses `shadow.overlay`; `Box`'s `boxShadow` prop is typed to the `shadow.level` namespace, so the surface takes `shadow.level.1` — same 8px blur and 0.25 alpha, but 1px of y-offset instead of 4px and a darker navy instead of a light periwinkle, so it reads slightly heavier than a menu's. Closing it means widening `Box.boxShadow` to the whole `shadow` namespace, which is a change to a shipped primitive and wants its own PR. Flagged in the spec.

### `<SelectButton />`, exported from `Select`

The field visual `Select` renders for its own trigger was unreachable outside the directory — `Select/index.ts` only re-exported `Select`, and `SelectTrigger` was welded to `useInternalSelectContext()`. Meanwhile the same visual had been independently re-copied in `MultiSelect`, `DatePicker`, and hand-rolled in `ForgeLayoutControls`.

It's now a public presentational component, with `SelectTrigger` refactored down to context plumbing over it. **`Select`'s own public API is unchanged**, and its existing test suite passes untouched — that's the regression check. Splitting `InputField`'s tokens out of its `root` mixin is what lets a field visual take the tokens without the column layout.

### Also in here

- **`fix(Box)`**, in its own commit: a `<Box />` nested inside a styled `<Box />` inherited its parent's `boxShadow`, `borderColor`, `borderStyle`, `zIndex`, `opacity`, `objectFit`, and `overscrollBehavior`, so a shadowed card painted its shadow again on every box within it. Found while building the popover surface. A `Nested` story is the guard, since jsdom applies no CSS-module stylesheets.
- **`fix(Select)`**: a placeholder longer than a narrow field wrapped to a second line and made the field taller than its neighbors. Placeholders now truncate, the way a selected value already did.

### Reviewing

The four content stories (`ShipmentPackaging`, `CustomPackage`, `SelectRate`, `Items`) are the acceptance test for whether the API is actually sufficient — they're built from the Figma frames using only design-system primitives. Worth walking:

- **`SelectRate`** and **`ScrollingContent`** exercise the `min-height: 0` chain. If it's wrong, the body won't scroll and the header/footer won't pin.
- **`SelectRate`** also validates `width="trigger"` — the panel should match the trigger on first paint with no flash.
- **`Components/Select`** — every existing story should look and behave identically, especially the descriptive two-line mode and all three sizes. That's where the `SelectTrigger` refactor could regress.

Two gaps the stories work around rather than invent past, both worth logging as follow-ups: there's no image/thumbnail component (`<Box as="img">` stands in) and no `Link` component (`Text`'s `as` union excludes `"a"`).

## ✅ Checklist

- [x] Specs are up-to-date
- [x] Changeset is added


ranebo (approved):
🔥

